PR #5, opened 2018-01-07
CF IDE mode setting was being written to PPIDE variable in RC2014 standard config

electrified:
Apologies if the indentation is off, used tabs but it still looks weird in the diff.

Ed


wwarthen:
Thanks for the fix Ed!

PR #6, opened 2018-01-27
Add support for DS1302 clock on RC2014

electrified:
Initial commit for RC2014 DS1302 clock support

Fairly arbitrarily picked port 0xC0, seems to work OK and not conflict with other devices although open to changing it :)

img_20180127_172623


wwarthen:
Sounds great Ed.  Board looks very nice by the way (very clean layout).

I will review and merge everything later today.

Thanks!

Wayne

PR #15, opened 2018-10-30
Move along, nothing to see here

b1ackmai1er:


b1ackmai1er:
Hi Wayne,

Is there a solution for assembling source in the hbios area which needs
more sophisticated macro capability?

Thanks

wwarthen:
Not really.  For standalone code, you could use a different assembler.
However, it would be a lot of work to port HBIOS to something else.

If you want to look into generic preprocessors, that might be an option.  I
don't particularly like adding additional tools to the build process though.

What problem are you trying to solve?

-Wayne

wwarthen:
I assume this is a standalone assembly.  As such, you may want to use a
different assembler.  See how BDOS/CCP are assembled in the CPM22
directory.  These use a real 8080 or Z80 based assembler running under
"zx".  Might make the whole porting effort easier.

-Wayne

b1ackmai1er:
Hi Wayne,

I have spent some time trying to convert a couple of Forth sources to TASM
but it didn't go well.

TASM has a tendency to stop without generating a PRN file so you can dubug
the assembly structures.

So I am going to go with your suggesting of using the cp/m emulator to
assemble the code.

Regards Phil

wwarthen:
The original selection of TASM for HBIOS predates my involvement in the
RetroComputer projects.  I believe it would have been better to use one of
the CP/M assemblers with the emulator for everything from the beginning.
For now HBIOS, is stuck with TASM.  However, almost anything new that I am
porting is done with a vintage assembler via the emulator.  I find it much
easier since the "porting" tends to be trivial.

Good luck.

Wayne

b1ackmai1er:
Thanks Wayne,

Forth is now assembling as part of the BuildShared script but probably
needs to be part of the HBIOS Build script but it is a good starting point.

The addition of the another assembler Z80MR and the Forth package add about
200Kb uncompressed to the package and the assembly process is very quick
under the emulator like the CP\M assembly.

If hperaza
<https://www.retrobrewcomputers.org/forum/index.php?t=usrinfo&id=241&>
releases the source code for his  ZSM4 assembler then it would good to
consider it as the default assembler.

I think the work that you and others have done on Retrobrew computers and
ROMWBW is pretty amazing.

Anyway, I'm off to do some gardening and hope to come back and finish this
off this afternoon/tonight.

Regards phil.

b1ackmai1er:
Hi Wayne

Forth is ready to try.

[image: image.png]

This could do with some real tidying up - and testing by a Forth user but
it is a starting point.

Did the pull request and then realized I hadn't include the assembler and
build script modification.

I have uploaded and think the pull request will include them. Lemme know if
these do not come through ...

[image: image.png]

Regards Phil.

PR #16, opened 2018-11-05
CamelForth ROMWBW

b1ackmai1er:
Feel free to change the build process on this. It is not very clean. Access to paths and standard definitions is difficult. Can you please confirm this builds in case I have missed uploading anything. Thanks.


wrljet:
Wayne,

I don't know if or where there is a correct place to discuss this stuff, but 
I just wanted to comment, remember typing in Fig Forth from a listing in 1979.

Bill

PR #18, opened 2018-11-07
Force DS1302 charging at reboot.

b1ackmai1er:
When using backplane supercap or battery, RTC memory is lost when changing EPROM on the SBC V2, so charging gets turned off, so eventually Supercap or battery runs down, so RAMF memory is lost.


wwarthen:
Note that setting the RTC into charging mode on a system that does not have a chargeable power source is dangerous.  Enabling force charging in EEPROM should be done only if you are sure of what you are doing.

PR #25, opened 2019-03-12
SIO & VDU updates

b1ackmai1er:
SIO fix.
VDU additional modes.
Working on getting back into sync so PIO driver can be committed.


wwarthen:
Thanks Phil!

PR #27, opened 2019-04-08
PIO driver

b1ackmai1er:
Base code for Zilog PIO driver.
Output using bit control, blind I/O works.
Not yet implemented are modes requiring interrupts.
Config updates for ECB-4PIO and ECB-Zilog Peripherals board.


wwarthen:
Thanks Phil.

PR #31, opened 2019-05-22
Display Z180 processor type at boot.

b1ackmai1er:
Can you please test on 180 platform. Thanks

### DIFF
--- a/Source/HBIOS/hbios.asm
+++ b/Source/HBIOS/hbios.asm
@@ -1255,6 +1255,7 @@ PSCNX	.EQU	$ + 1
 	PRTS("MHz$")
 #IF ((PLATFORM == PLT_N8) | (PLATFORM == PLT_MK4) | (PLATFORM == PLT_RCZ180))
 	LD	L,1			; flag Z180
+	LD	D,0
 	OUT0	(Z180_ASTC1L),D		; D = 0 at this point
 	IN0	A,(Z180_FRC)		; supposedly only on S-class
 	INC	A			; FF or 00 -> 0 or 1 (weak S-class or higher)
--- a/Source/HBIOS/hbios.asm
+++ b/Source/HBIOS/hbios.asm
@@ -1254,8 +1254,20 @@ PSCNX	.EQU	$ + 1
 	CALL	PRTD3M			; PRINT AS DECIMAL WITH 3 DIGIT MANTISSA
 	PRTS("MHz$")
 #IF ((PLATFORM == PLT_N8) | (PLATFORM == PLT_MK4) | (PLATFORM == PLT_RCZ180))
-	CALL	PC_COMMA
-	PRTS(" IO=0x$")
+	LD	L,1			; flag Z180
+	OUT0	(Z180_ASTC1L),D		; D = 0 at this point
+	IN0	A,(Z180_FRC)		; supposedly only on S-class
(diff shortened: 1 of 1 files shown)


wwarthen:
Thanks Phil

MorfeoMatrixx:
Phil, this should be updated to include the new **SCZ180** platform.
Sorry I can't put a PR myself now because I'm still learning how to use Github : -0

Cheers, JL.

feilipu:
Not the Phil from this commit, but anyway the commit relates to the processor type, not to the platform. So it is correct as it stands.
Cheers, @feilipu

MorfeoMatrixx:
Hi Phil, sorry for my blatant incompetence, but in my SC126 board (SCZ180 family, uses Z180 CPU also) the Z180 processor (sub)type is not displayed. Is this still correct ?

Cheers, JL

feilipu:
For  the record, I see this.

```bash
RetroBrew HBIOS v2.9.2-pre.20, 2019-11-15

SC130 Z8S180-N @ 36.864MHz IO=0xC0
1 MEM W/S, 2 I/O W/S, INT MODE 2
512KB ROM, 512KB RAM

~~~snip~~~
```
The differences from standard are because I'm configuring `Z180_CLKDIV = 2` and one memory wait state (the flash / RAM is too slow for 0 wait states) `Z180_MEMWAIT = 1`. But you should see in essence exactly the same thing.

MorfeoMatrixx:
Thanks for checking, I can’t test it now but tomorrow I’ll get back to you. I see in your message that the “Type:” string introduced in the other Phil’s patch is missing, this is what I saw with my board.
I didn’t review the current method code.

MorfeoMatrixx:
Phil, this is what I get with the stock ROM...

  RetroBrew HBIOS v2.9.2-pre.1, 2019-07-23

  RC2014 Z8S180-N @ 18.428MHz IO=0xC0
  0 MEM W/S, 2 I/O W/S, INT MODE 2
  512KB ROM, 512KB RAM

 I know it's not the last pre-release, but when trying to flash the newer code using the FLASH4 CP/M program. I get a message saying that the chip is not compatible (I reported this to the author, Will at sowerbutts dot com).

With the newer code (that I didn't look at...) evrything seems fine, it is reported as "-N" as it should. The old method used appended a "Type;" string.

So at least  I was right when in my previous mail I mentioned my "blatant incompetence"... ;-)

Cheers, JL.

wwarthen:
> I know it's not the last pre-release, but when trying to flash the newer code using the FLASH4 CP/M program. I get a message saying that the chip is not compatible (I reported this to the author, Will at sowerbutts dot com).

This is usually because the write protect jumper is shorted on your board.  Strangely, shorting the W/P jumper causes the chip identification process to indicate a different (and unknown) chip type.

MorfeoMatrixx:
Wonderful ! Many thanks for your support... the more I tinker with your
board, the more I love it !!!

Now I will try to program the new version with my serial baudrate update in
a new chip.

Cheers, JL.

________________________________________________________
B>flash read j:romwbw00.bin
FLASH4 by Will Sowerbutts <will@sowerbutts.com> version 1.2.3

Using RomWBW (v2.6+) bank switching.
Flash memory chip ID is 0xBFB7: 39F040
Flash memory has 128 sectors of 4096 bytes, total 512KB
Read complete.

PR #34, opened 2019-05-28
Add debugging module mio

b1ackmai1er:
Cached debug out.
Hi Wayne, I was struggling with seeing what was going on in the pio pre-init code due to lack of ability to display debug output. So this little enhancement adds a driver so normal output routines can be used to capture debug info by outputting to memory 'cache'. Once the normal i/o is initialized, the 'cached' output then gets displayed. I've added some demo debugging code which will have to be taken out once you review and approve. As is, the only change your will see is a hex dump of the IVT table prior to the boot message. But if you assemble with WBWDEBUG .SET USEMIO then when booting the IVT table will be displayed from prior to preinit, and a regdump will be displayed prior to and post preinit and then the current IVT table from after all initialization is complete. Will email you a screen shot to clarify. Limitations: $8000 is used as the memory cache. In app mode this seems to overwrite ZCPM. But CPM works. If you output enough stuff  you will probably overwrite CPM and HBIOS too i assume! No limit set at the moment. Perhaps you can recommend better starting point then MEMTOP. Works better in ROM mode. Anyway let me know what you think. I like it but you have better understanding on how this fits in with everything else.


wwarthen:
Accepting these changes, but I will soon be making a few supplemental changes to ensure that this does not break anything when it is not enabled.

PR #35, opened 2019-06-10
 Invalid BIOS (signature missing) 

b1ackmai1er:


wwarthen:
Really good catch on this one Phil!  I have definitely heard of occasional issues like this and I think you found the source of the problem.  Thanks!!!

wwarthen:
Hi Phil,

This one was a really good catch.  Thank you very much!

Wayne

PR #38, opened 2019-08-04
usb-fifo driver et al

b1ackmai1er:
Add ecb-usb-fifo driver.
Change the driver pre-init list to same structure as init list, to facilitate the following:
Add ability to force which CIO driver runs first so default console can be changed without manually shifting driver sequence.
Update to PIO driver, still work in progress.


wwarthen:
Thanks Phil!

PR #42, opened 2019-08-24
DS1210 Battery Low message on boot

b1ackmai1er:
Displays BATTERY LOW as part of the DSRTC initialization if DS1210 detects battery voltage below 2V


wwarthen:
Thanks Phil.

PR #44, opened 2019-09-01
Refactor of device summary

b1ackmai1er:
Hi Wayne,

Attached changes improve code readability albeit with some processing overhead which I think is ok since all the initialization code is execute once only. With the changes I think there is an overall space saving too.

Regards Phil


wwarthen:
This is great.  Thanks Phil.

wwarthen:
I really like this Phil.  Thank you!

PR #48, opened 2019-10-01
Change status LED use hbios.asm for SC126/130

sccousins:
This modification changes the function of the status LED on SC126/130 such that the LED is turned off when the user selects an option from the startup menu and then flashes on during disk activity. If LEDDISKIO is FALSE the LED remains on after the menu selection.


wwarthen:
Thanks Steve!

Issue #53, opened 2019-11-07
hbios - z180 real time unix time - (enhancement)


feilipu:
I'd like to interface the HBIOS TIMER functions with the [Y2K](http://www.nongnu.org/avr-libc/user-manual/group__avr__time.html) epoch (related to Unix Epoch). Whilst it is possible as stands, would it be possible to suggest some minor changes to the HBIOS TICKS implementation (on Z180)? These suggestions would make things much simpler.

If the suggestion is approved, I'll do a PR to suit.

1. The Y2K Epoch is 32 bits, with the tick increment per second. -> Make the timer location 4+1 bytes, with the lowest byte measuring fractional seconds. Unix Epoch can then be loaded from RTC into the 4 bytes second counter with standard conversion code.

2. Match the `PRT0` tick rate to a binary multiple (i.e. 256Hz or 128Hz) and have the fractional second (1 byte) counter roll over to increment the second counter (4 bytes). -> Keeps the interrupt handling very simple (as it is now). Keep translation of [fractional seconds to nanoeconds](https://github.com/z88dk/z88dk/blob/master/libsrc/_DEVELOPMENT/target/scz180/driver/time/z180/clock_gettime.asm) out of time critical code.

The outcome is to allow the z88dk [`sys/time.h`](https://github.com/z88dk/z88dk/blob/master/include/_DEVELOPMENT/sdcc/sys/time.h) unix like functions to easily connect into the HBIOS API.

Thoughts?

wwarthen:
Hi Phillip,

I do like the idea of supporting a "seconds" counter.  There are many
instances where this would be useful.  However, the existing 50Hz tick
counter is very specifically intended to be optimal for MP/M and audio
playback.

I think it would be pretty straightforward to maintain a separate 32-bit
counter that just counts seconds.  Essentially, the seconds counter would
be incremented 50 ticks.  I would need to add something to set and query
the seconds counter, but that is not hard.  There would be very little
overhead to maintaining the additional counter because 49 interrupts out of
50 would do almost nothing extra.

Would this be a reasonable compromise?

By the way, I want to make sure you know that not all platforms running
RomWBW have timer hardware.

Thanks,

Wayne

feilipu:
Wayne,

yes, if 50Hz is a critical timer then it makes a very good compromise to have a 50Hz fractional second counter feeding into a 4 byte Y2K Epoch second counter. 👍 

[...]

PR #54, opened 2019-11-09
hbios - system ticks

feilipu:
Relating to #53, an enhancement to get an intrinsic time stamp in _seconds.ticks_ with 50Hz resolution.

Through #53 changes, the `INC32HL` function is now called twice now in the interrupt, including once per 20ms.

So suggest to change function definition so that `HL` is not preserved. Checked throughout the repository, and every other call to `INC32HL` does not require the `HL` to be preserved, and in fact they all immediately overwrite `HL`.

OLD
``` asm
INC32HL: ; minimum impact 88 ticks, total 14 bytes
	PUSH	HL  ; 11 - 1
	PUSH	BC ; 11 - 1
	LD	B,4 ; 7 - 2
INC32HL1:
	INC	(HL) ; 11 - 1
	INC	HL ; 6 - 1
	JR	NZ,INC32HL2 ; 12/7 - 2
	DJNZ	INC32HL1 ; 13/8 - 2
INC32HL2:
	POP	BC ; 10 - 1
	POP	HL ; 10 - 1
	RET ; 10 - 1
```
NEW
``` asm
INC32HL: ; minimum impact 22 ticks, total 11 bytes
	INC	(HL) ; 11 - 1
	RET	NZ ; 11/5 - 1
	INC	HL ; 6 - 1
	INC	(HL) ; 11 - 1
	RET	NZ ; 11/5 - 1
	INC	HL ; 6 - 1
	INC	(HL) ; 11 - 1
	RET	NZ ; 11/5 - 1
	INC	HL ; 6 - 1
	INC	(HL) ; 11 - 1
	RET ; 10 - 1
```

Added a `TICKSPERSEC` configuration item, to avoid magic numbers appearing in the code.


wwarthen:
Thank you Phillip.

PR #57, opened 2019-11-14
hbios - z180 - tidy prt0 setting

feilipu:
Tidy up PRT0 setting to save save a few ticks, and make the code more readable.

``` asm
;   LD      HL,CPUOSC/TICKSPERSEC/20-1    ; 50HZ = 18432000 / 50 / 20
    LD      HL,(CPUKHZ)         ; 50HZ = 18432000 / 20 / 50 / X, SO X = CPU KHZ  ; 3 - 16
    LD      B,0                 ; 16-BIT I/O ADDRESS                        ; 2 -  7
    LD      C,Z180_TMDR0L       ; INITIALIZE TIMER 0 DATA REGISTER          ; 2 -  7
    OUT     (C),L                                                           ; 2 - 12
    INC     C                                                               ; 1 -  4
    OUT     (C),H                                                           ; 2 - 12
    DEC     HL                  ; RELOAD OCCURS *AFTER* ZERO                ; 1 - 6               
    LD      C,Z180_RLDR0L       ; INITIALIZE TIMER 0 RELOAD REGISTER        ; 2 -  7
    OUT     (C),L                                                           ; 2 - 12
    INC     C                                                               ; 1 -  4
    OUT     (C),H                                                           ; 2 - 12
    LD      A,%00010001        ; ENABLE TIMER0 INT AND DOWN COUNTING        ; 2 -  7
    OUT0    (Z180_TCR),A                                                    ; 3 - 13
                                                                            --------
                                                                             22 -119
```

```asm

;   LD      HL,CPUOSC/TICKSPERSEC/20-1    ; 50HZ = 18432000 / 50 / 20
    LD      HL,(CPUKHZ)         ; 50HZ = 18432000 / 20 / 50 / X, SO X = CPU KHZ  ; 3 - 16
    DEC     HL                  ; RELOAD OCCURS *AFTER* ZERO                ; 1 - 6
    OUT0    (Z180_TMDR0L),L     ; INITIALIZE TIMER 0 DATA REGISTER          ; 3 - 13
    OUT0    (Z180_TMDR0H),H                                                 ; 3 - 13
    OUT0    (Z180_RLDR0L),L     ; INITIALIZE TIMER 0 RELOAD REGISTER        ; 3 - 13
    OUT0    (Z180_RLDR0H),H                                                 ; 3 - 13
    LD      A,%00010001         ; ENABLE TIMER0 INT AND DOWN COUNTING       ; 2 -  7
    OUT0    (Z180_TCR),A                                                    ; 3 - 13
                                                                            --------
                                                                             21 - 94
```


wwarthen:
Excellent!  I keep forgetting that OUT0 on Z180 can take any register.  So much cleaner!

wwarthen:
Note that you will see one small change to your work at my next check in.  It is an odd thing, but if you look at the Z180 doc on the reload timer, it appears that only TLDR is one less than the desired count.  Looks like TMDR should be set to the actual desired count.  I think this is because the actual interrupt occurs on transition from 1 -> 0.  So I moved the DEC HL to after the setup of TMDR.  In the end, it it really doesn't have any practical impact, so just a nit.

feilipu:
Best to have it right.

Related, I ran a short RTC test before retiring yesterday and found the tick "clock" is running quite fast. Didn't analyse enough to fix. Will have another look this weekend.

I'm running `Z180_CLKDIV 2` and `Z180_MEMWAIT 1`, so perhaps there's something in the `CPUKHZ` calculation that doesn't allow for these configurations?

PR #58, opened 2019-11-15
hbios - z180 CMR/CCR config

feilipu:
The cpukhz calculation for z180 seems to be off when the `Z180_CLKDIV = 2` multiplier is selected.

I believe it is because [at this point the A](https://github.com/wwarthen/RomWBW/blob/master/Source/HBIOS/hbios.asm#L1285) register is not filled properly with `	LD	A,(HB_CPUTYPE)	; GET CPU TYPE` but rather contains `$80` from the `LD	A,$80` previous instruction.

This PR fixes the issue, and removed double code by only testing/setting the multipliers once, and setting CMR before CCR.

~~BUT, this in-turn creates another issue that the `HB_CPUSPD` calculation misreports the actual CPU speed. Not handled by this PR - make sure I don't break things for others.~~

__EDIT__ fixed the configuration logic. The `HB_CPUSPD` calculation is correct. Doh!

PLS - Phillip


wwarthen:
Thank you for this.  No time right now, but will review tomorrow and merge,

feilipu:
Wayne,
I'm still missing a trick here. The PRT seconds counter is now running about 4x fast when using the `Z180_CLKDIV = 2` configuration. I haven't found why.

wwarthen:
In your updates, you are using "=" to test for equality in the conditional assembly tests.  I think that TASM requires "==" to test for equality.

Although it does not explain your problem, I also see I have an order of execution issue.  The PRT0 setup is done before the CPU speed setup.  So, the CB_CPUKHZ variable may not have an accurate value in it at that point.  Additionally, the PRT0 setup code is using CPUKHZ.  It should use CB_CPUKHZ.

feilipu:
I've fixed the equality statements, but that is probably not the end of it.
Perhaps best to leave it to you to sort out the execution order issue.
P.

wwarthen:
Thanks Phil.  I will do some further review/testing.

-Wayne

wwarthen:
Just checked in a small set of changes that I hope resolve all outstanding issues with CMR/CCR configuration.  Note that I did wind up adding the double code.  This is because without it, if the user has a Rev. K CPU, then cpu speed would revert to 1/2x instead of 1x.  I didn't see a better way to handle this.

feilipu:
Thanks. I'll test later on today.

The[ note about setting CMR before CCR](https://github.com/wwarthen/RomWBW/pull/58/files#diff-3791254867c3179e9426291f9bfdf980L1289) arose from [this discussion at RetroBrew](https://www.retrobrewcomputers.org/forum/index.php?t=msg&th=316&#msg_5045). I didn't really see the issue in my code because I wasn't adjusting the CCR in YAZ180, but I've followed the guidance.

It might be good just to set CCR again after the CMR setting.

P.

wwarthen:
> The[ note about setting CMR before CCR](https://github.com/wwarthen/RomWBW/pull/58/files#diff-3791254867c3179e9426291f9bfdf980L1289) arose from [this discussion at RetroBrew](https://www.retrobrewcomputers.org/forum/index.php?t=msg&th=316&#msg_5045). I didn't really see the issue in my code because I wasn't adjusting the CCR in YAZ180, but I've followed the guidance. It might be good just to set CCR again after the CMR setting.

Got it.  I will adjust the code.

PR #61, opened 2019-11-30
[hbios] dbgmon hxload rework

feilipu:
I've been having some trouble with the Intel HEX loading function in the dbgmon.

- Firstly, sometimes junk in the serial port queue caused transfers to die before the first character was successfully read. Said the other way around, the first character transmitted was required to be a colon, and often it wasn't.
- Secondly, one word... slow.

I've done a bit of a rework on the function, and it resolves the issues I was facing. Now I'm not getting issues with junk in the serial queue before the first colon. And, it can now keep up with line rate at 115200 baud off the z180 SC130. For the RC2014 z80 target at 115200 baud there is still a small inter-character delay required. There are too many `push`/`pops` and bank switches necessary per character to remove this requirement, but at least the inter-character delay required is substantially lessened.

The only down side is that the hex loading function can't be interrupted when it is waiting for a colon. The remedy is a quick reboot.


wwarthen:
Thanks Phil.

I am tied up today, but will review everything tomorrow and merge your changes.

-Wayne

wwarthen:
I have not tested thoroughly, but this looks good and I have merged the changes.

I suspect that you do not have hardware flow control working (which many people do not).  I have never seen problem with losing characters or garbage when flow control is in effect.  Regardless, these changes are very helpful.

Thanks!

PR #62, opened 2019-12-01
Tune defaults to .PT3 file type if none given

b1ackmai1er:


wwarthen:
Thank you Phil.

Is it common for tune files to have no extension?

-Wayne

b1ackmai1er:
Hi Wayne,

I havn't seen one yet, and the pt3 file format appears to be the most
common format, so I think this is a safe assumption to make.

Regards Phil.

PR #64, opened 2020-01-07
HBIOS - sd.asm efficient byte mirror

feilipu:
Use the efficient byte mirror routine Using the efficient mirroring routine, from John Metcalf.

http://www.retroprogramming.com/2014/01/fast-z80-bit-reversal.html


wwarthen:
Thank you Phillip!

PR #66, opened 2020-01-10
Add 80x43 video mode to VGA3 driver

b1ackmai1er:
Do not have a CVDU card so was unable to test changes do not affect it but it assembles ok.


wwarthen:
Thanks Phil!

PR #68, opened 2020-01-15
Initial compressed fonts support

b1ackmai1er:
Includes fix for VCDU driver I think. CVDU not tested.


wwarthen:
Wow, there is a lot here Phil.  The lzsa compression is very cool and may have uses outside of the font data.  I am going to go ahead and merge all of this, but will be pretty much untested by me.  I will try to do some testing over the next few days.

Thanks!

PR #69, opened 2020-01-17
Implement "safe mode" boot process

b1ackmai1er:
Lots of ways this could be done. I have chosen to set it up so it just boots with a minimum driver set.

On SBC test  by setting BT_REC_TYPE to BT_REC_SBC02 and use the 1-bit jumper (crtact) jumper to enable.

Options for other platforms needs to be considered..


wwarthen:
Thanks Phil.

PR #71, opened 2020-01-29
ECB Cassette and Recovery Mode change

b1ackmai1er:
Create an equate for changing baud rate through custom config rather than the driver. Add recovery mode option to use Ring indicator line on SBC UART.


wwarthen:
Looks good, thanks Phil.

PR #74, opened 2020-02-05
ppide update to begin support for multiple interfaces.

b1ackmai1er:
Hi Wayne, can you check for breakage. Appears to work ok in standard mode for me.  Does recognize my flash drive on my SBCV2 and works fine and recognizes my HD on my DIO V3 but gives sector errors. Lots of stuff around the device numbering is confusing to me.  Regards Phil


wwarthen:
Hi Phil,

This is a sizeable change, so I definitely need to take some time to get comfortable with everything.  I will work on it, but may take me a few days.

Thanks,

Wayne

b1ackmai1er:
I missed uploading std.asm which includes USEZ80OPT .EQU FALSE ; USE UNOFFICIAL OP CODES
definition. OUT (O),0 did not work for me. Am going to run zexddoc on my machine to see what it says. Not sure what your policy is on unofficial opcodes.

feilipu:
It doesn’t work on Z180. 

If you’re going to use it please wrap with conditional based on CPU.

wwarthen:
Yes, I was just about to point that out when you beat me to it!  I am fine with use of undocumented instructions, but when they are not available on all CPUs, I avoid them unless there is a strong reason to use it.  In the place where you have used it is not performance sensitive, so I don't think it is warranted in this specific case.

wwarthen:
Thanks Phil!

PR #76, opened 2020-02-09
Update romldr to support more than 9 drives.

b1ackmai1er:
Update menu and input system to support more than nine drives. Drives over 9 should continue to be listed as 10, 11, 12 etc. This means that when there are more than 9 characters, 2 characters are required to input the drive to boot. For example '1' '4' would be required to be entered to boot drive 14. '0' '2' would be required to boot drive 2, or alternately '2' <enter> would boot drive 2. If there are less than 10 drives the the one character menu system is retained. I have made allowances for dsky but it does not work on my system as I don't have one, so it may need a few tweaks, Also don't have more than 9 drives so I couldn't confirm it truly works but debugging output looked good. Autoboot modes still work as before but do not support more than the first nine.


b1ackmai1er:
Above should read  2 ENTER will boot drive two.

wwarthen:
Thank Phil.

Ultimately, I think this should be more of a simple command line interface.  That would allow boot parameters to be passed to the boot target.  For now, this is just fine.

b1ackmai1er:
Cool, are you thinking of something along the lines of linux lilo boot
loader.?

Did you have a chance to check this out with dsky?

Are you happy with the inclusion of the game? Please let me know if it
doesn't suit the style of romwbw that you want.

Best wishes Phil.

wwarthen:
Yes, something along the lines of lilo, though not nearly as sophisticated.

I have not tried DSKY yet.  Sorry.  I'm about half way through propagating the PPIDE changes into IDE, so I want to finish that first.

Sure, the inclusion of a ROM game is great.  I don't want to start including a suite of ROM games,but having one in ROM is a nice.

Thanks!

Wayne

PR #77, opened 2020-02-10
Add a user rom pack template to romldr

b1ackmai1er:
HI Wayne, What do you think about this idea. I've added a "USER ROM" option to romldr. Effectively it is a prepopulated placeholder for a user to add their own code and have it as part of the romldr without the need to modify build files, or worry about having to reintegrate anything with new ROMWBW releases. Ive set this up as 4K at the end of the rom bank, but size could be bigger or smaller. The placeholder file could be  populated with sample code for accessing HBIOS functions, I've got a few other ideas around this but didn't want to go too far in case you didn't like the idea. I think the big plus with this is that it is an easy entry point for new users to do some system coding.  Happy to revert, it's just an idea. Regards Phil


wwarthen:
I really have not had time to look at this, but it seems straightforward, so I am going to merge it.

Thanks!

Wayne

PR #78, opened 2020-02-14
Update cbios.asm 

b1ackmai1er:
Change RAMF to 512 directory entries as 4Mb drive space cannot be effectively used without enough directory entries and extents to hold a large quantity of small files or a small qty of large files.


wwarthen:
Thanks Phil.

For the record, I would love to change the number of directory entries for hard disks to 1024 or 2048.  Unfortunately, that would break all existing media.  Sigh.

b1ackmai1er:
Why don't you handle this the same way other software transitions are made:

Identify anything that you want to change that breaks existing
installations and call that the ROMWBWN (New). Then have the existing build
ROMWBWL (Legacy).

For all new installers, direct them to ROMWBWN.
For existing users, direct them to ROMWBWL.

Continue parallel build for a period of time with the only real difference
being a configuration file.

At some point in time, like a major release, move to releasing only ROMWBWN
with Legacy support only as a custom build option.

Regards Phil.

wwarthen:
Yes, that sounds like the right approach.  I need to come up with a list of breaking changes that I want to make and bundle them together.  I am sure there were more, but I don't remember right now.  I need to keep a list as they occur to me.

In addition to the "legacy" build option, I think I would introduce this as version 3.x.  In theory, breaking changes should increment a higher node of the version.  I could release 3.x which defaults to the new configuration, but support a legacy switch initially.

Thanks for your thoughts.

-Wayne

PR #81, opened 2020-02-19
clean unix build after merge

cm68:
This pull fixes the few discrepancies in the build, as well as compile warnings in the linux cross tools.


wwarthen:
Thanks Curt!

PR #82, opened 2020-02-21
clean build, bit-identical.

cm68:
I also added a top-level Readme.unix file

it passes a regression test.  I did fix a few bugs.  there was a .DS that should have been a .FILL and the
aforementioned string problem.


wwarthen:
Thanks Curt!

PR #88, opened 2020-02-26
uz80as expression parser fix for .db ~(1|2) != 0xfe

cm68:
I also added the ability to build with the same version number as extracted from a previous build that is pointed at for diffing.


wwarthen:
Thanks Curt!

PR #93, opened 2020-03-01
Nascom manual update

b1ackmai1er:
Full text version recreated from original scan. From Jan S RC2014  google group.


wwarthen:
Thanks Phil!

PR #94, opened 2020-03-01
Dean/linux tweaks

vipoo:
Some minor tweaks/fixes for linux pipeline

1. Added a gitignore - was a bit of a challenge to confirm what files are generated, and what are just part of the repo.

2. the change for build.sh is only to make the bin excutable under linux (eg ./Build.sh) - there is also a minor whitespace change - no actual change to content

3. And in my hacking with the code, noticed the issue with the HBIOS/build.sh script not handling missing files very well - i think i have understood the intent of that code - but i may have it wrong


wwarthen:
Hi Dean,

Thanks for being a new contributor to RomWBW!  Changes look good.  Just a couple comments.

In the distant past, attempts to use .gitignore became problematic because it confused some users and it kept getting out of sync with the reset of the build.  Curt wanted to add .gitignore files and I asked him to remove them.  Since you are the second person that expects them, I will accept them for now and see how it goes.

I'm not sure about your comment about needing to handle missing files in the disk image creation process.  I don't think there should be missing files, but I certainly could have missed something  What files did you see missing?

Thanks!

Wayne

wwarthen:
Sorry, one more thing.  Binary/Apps/Tunes/ReadMe.txt was actually supposed to be there.  Removing them was actually a problem in clean that I think is now corrected.  I will be putting that ReadMe.txt back.

Thanks.

vipoo:
Hi Wayne,

I am glad my little contribution was helpful - I feel like I am only understanding a small percentage at the moment - so its quite possible i have assumed wrong things .

In regards to the skipping files - i stumbled on this , as part of hacking with the project - noticed there was this code:

```
f=$(../../Tools/unix/casefn.sh ../../Binary/Apps/$i.com)
if [ -z "$f" ] ; then
	echo " " $i "not found"			echo " " $i "not found"
else
```

problem was the if statement would never have a chance to test for the missing file, as the previous line would bom out.  This was due to the script running with the -e option (stop on error).  So I wrapped the casefn call, with (set +x, set -x) to allow it to proceed -  and give a clean message of the issue.

wwarthen:
Hi Dean,

OK, I got it now.  I had thought you found actual missing files, but I now understand you are just making the code consistent.

Thanks!

Wayne

PR #95, opened 2020-03-01
Make .gitignores for Tools/unix more specific

electrified:
The new .gitignore blanket excluded the whole of the `Tools/unix` dir. This made it harder to add new files.

Have removed the global ignore and added specific files.

I contemplated creating individual .gitignore files in the specific tool directories, but have left it all in one for the time being.

Also added the macOS output dir.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,7 +55,21 @@ Source/ZPM3/gencpm.com
 Source/ZPM3/gencpm.dat
 
 Tools/Linux
-Tools/unix
+Tools/Darwin
+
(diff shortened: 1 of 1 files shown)


wwarthen:
Thanks Ed.  I have really mixed feelings about the use of .gitignore in general.  In my experience, they are never quite right and cause problems.  Several folks seem to want them though.  So, we'll see how it goes.

wwarthen:
Oh, and I would prefer not to put .gitignore files in all the directories for now.  Just seems like even more work to maintain.  Thanks.

Issue #87, opened 2020-02-24
hbios - access to z180 interrupt table - freertos


feilipu:
I'm interested to use the PRT1 in the z180 to generate regular timing interrupts for my program. But I can't find in the documentation where the z180 interrupt address table is stored, or identify which (if any) HBIOS call is used to insert an interrupt handler into the z180 interrupt address table.

An alternative (perhaps supported?) would be to have the regular HBIOS tick (50Hz) call a function in my program on every tick. This would save consuming the PRT1 for essentially the same function that is provided by the HBIOS tick.

Any guidance on the best way to provide a timing interrupt within HBIOS?
Or, without breaking HBIOS?

wwarthen:
It is entirely possible to hook the existing Z180 timer interrupt in RomWBW via API calls.  I will provide some better instructions as soon as I have a bit more time.

wwarthen:
There is an API in RomWBW HBIOS for interrupt hooking.  It is documented (albeit poorly) in the RomWBW Architecture document.  It is the SYSINT function ($FC).  This is at the very end of the document.

All of the standard Z180 configs are IM2 based.  If you look at hbios.asm, search for HBX_IVT.  This is the interrupt vector table.  Just above it is a comment that documents the interrupt slots assigned to the standard interrupts.  In your case, you are going to hook slot 2 (TIM0) which is used for all Z180's as a 50Hz timer.

There is a good example of hooking the vectors in the inttest.asm application included with RomWBW.  The app is more generic than you need, but I think you will easily see the simple routines for hooking and unhooking interrupt handlers.

The new interrupt handler **must** be in the top 32K of memory.

I'm hoping this is enough to get you started.  Let me know how you do and ask questions as needed.

Thanks,

Wayne

feilipu:
Thanks. I've had a look around in the code. From what I see the Z180 [address table entries](https://github.com/wwarthen/RomWBW/blob/master/Source/HBIOS/hbios.asm#L536) are all pointing at a [`HBX_INT`](https://github.com/wwarthen/RomWBW/blob/master/Source/HBIOS/hbios.asm#L588) with a following code byte to route the interrupt onwards viat the [`HB_IVT`](https://github.com/wwarthen/RomWBW/blob/master/Source/HBIOS/hbios.asm#L729).

[...]

PR #96, opened 2020-03-02
HBIOS - relocate HBX_TMPSTK

feilipu:
As discussed on #87, move the `HBX_TMPSTK` to an alternative slack space below `0xFF80`.

This provides more unencumbered space for the `HBX_STK`  (growing down from `0xFF00`) used when handling interrupts.


feilipu:
I couldn't think of a way to preserve the interrupt status, and still return AF (flags intact) from the function invoked, other than the sledgehammer of `HB_DI` & `HB_EI`. But that may be an acceptable compromise?

wwarthen:
I am fine with the relocation of HBX_TMPSTK -- good change.

However, please see my comments about the use of DI/EI in HB_INVOKE in issue #87.

Thanks,

Wayne

feilipu:
Closed because a better solution was found for #87.

PR #99, opened 2020-03-03
hbios - tidy TMPSTK

feilipu:
Complete the move of HBX_TMPSTK into the HBX_BUF space.
This relocation leaves the interrupt stack clear.

Noting that the HBX_BNKCPY uses HBX_BUF, so its temporary stack needs to exist in the space below HBX_BUF.

This is also useful because all HBIOS PROXY stacks will now reside in the `0xFF00` page, making testing for their use easy. Discussed in #87.


wwarthen:
Excellent, thanks!

wwarthen:
Hi Phillip,

I was getting sort of confused with the number and naming of the buffers/stacks.  I made another pass at it.  I don't think I changed anything functionally, but I find this to be a little easier to understand.  It is probably just me.  Anyway, let me know if you have any issues with the latest hbios.asm check-in.

Thanks,

Wayne

PR #100, opened 2020-03-04
hbios - wrap hbx_bnkcpy hbx_peek hbx_poke in di/ei

feilipu:
Because `HB_BNKCPY` is called in quite a few places across the code base, it isn't practical to protect it everywhere. So do the `DI`/`EI` inside the `HBX_BNKCPY` function, and do it once only.

__EDIT__ Also moved the `HBX_PEEK` & `HBX_POKE`  `DI`/`EI`guards inside their functions, because it is easier to maintain when cause & effect are in same function, saves a few bytes, and reduces cycles in DI.


wwarthen:
Hi Phillip,

Thanks, you are right about HBX_BNKCOPY being called directly from other places.  I am going to merge your changes, but it introduces a new problem.  The additional code reduces the size of the INTSTK down to 16 bytes for the N8 platform.  That is too small.

I am going to move the DI/EI protection for PEEK and POKE back into the main HBIOS code.  PEEK and POKE are not accessible directly, so this should be OK.  This will recover 9 bytes for INTSTK, so the N8 platform will have a 25 byte INTSTK.  Still far smaller than I would like, but should be sufficient.

I will check in this modification in a couple hours.

Secondly, I want to confirm how you are checking that HBIOS is active.  The first thing that HBX_INVOKE does is save the callers SP.  If an interrupt fires immediately after this and RTOS does a context switch, the new context could call HBX_INVOKE and clobber the SP value stored by the original context, right?  It may be necessary to set an explicit "HBIOS_ACT" flag as the very first step of HBX_INVOKE and FreeRTOS would need to check that to determine if it should back off.  Thoughts?

Thanks,

Wayne

PR #102, opened 2020-03-07
hbios - tmp stack tweak

feilipu:
There is a small chance that the context could be swapped invalidating the `HB_INVBNK` value,

This fix mitigates that chance by moving the stack pointer into the HBIOS page before touching `HB_INVBNK`, and no context swaps are attempted when the stack pointer in the `0xFF00` page.

And, updated the date.


wwarthen:
Thank you Phillip!

PR #104, opened 2020-03-12
hbios - remove mutex special files

feilipu:
I think there's little value in keeping special mutex enabled builds. It is a fair assumption that if people are keeping up with the pre-releases then they have the capability to build their own ROM.

This PR just removes the special files and sets the `HBIOS_MUTEX` to false for SCZ180 platforms, as a reminder that it exists.

I think I might change the comment wording too.

__EDIT__ Done. But not sure that it is more clear that way either.
Adding the mutex is more about managing reentrant calls, than "enabling" or "preventing" reentrant calls. The mutex blocks rentrant calls, but actually enables them. Not sure how this is best described. `PREVENT` or `ENABLE` both seem wrongish.


wwarthen:
Agreed.  Thanks Phillip.

PR #108, opened 2020-03-27
ReadMe.md

feilipu:
Remove poorly placed `\` preventing markup working.


wwarthen:
Thanks Phillip.

By the way, you may notice I checked in overtop of these files.  This is because these files are now actually generated files.  The generation is not part of the build process because it requires pp, pandoc, and MiKTeX which I'm not sure I want to embed in the build process.

Thanks, Wayne

PR #111, opened 2020-04-21
Dean/dev/general tooling

vipoo:
Hi Wayne,

This is a small PR, in prep for the sound driver changes - coming in the next PR.

Main changes are:

#### 1. I updated the linux make build to support optional ENV variables - to target a specific ROM images.

This allows the targeted builds in much the same was the the windows command line build script supports.
  
The env variables are:

  `ROM_PLATFORM`: - to speficy a ROM build - eg : RCZ80
  `ROM_CONFIG`: - to specify a config - defaults to std

If `ROM_PLATFORM` is not set, then it does as it previously did - and builds all images.

#### 2. Update sub makefile references to use `$(MAKE)`

There are uses of the following pattern:
  `cd <dir>; make`

I have changed them to use the make env pass thru:
  `$(MAKE) --directory <dir>`

This allows for some make switches to pass thru - eg -B.  And also helps if parallel builds are activated - it be nice to get the Makefile to fully utilise parallel builds....

#### 4. Whitespaces!
I am sometimes my own worst enemy - I have my development tools setup to adhere to some strict work patterns - its how I have encouraged the teams I work with.  

By avoiding trailing spaces - it reduces noise in commits/merges.
Tell me to stop it if you want - I dont wont to come across as trying to demand a specific and pedantic coding style.

#### 5. Driver function table check - hard error

I add a hard `FAIL` command within the preprocessor code that check the drivers have implemented the required count of functions.  I found sometimes i had got it wrong on the sound driver - and never noticed the failure - now if the count is not as expected - the assembly will fail.

So
```
#IF (($ - XXX_FNTBL) != (XXX_FNCNT * 2))
	.ECHO	"*** INVALID XXX FUNCTION TABLE ***\n"
#ENDIF
```

becomes
```
#IF (($ - XXX_FNTBL) != (XXX_FNCNT * 2))
	.ECHO	"*** INVALID XXX FUNCTION TABLE ***\n"
	FAIL
#ENDIF
```


wwarthen:
> This is a small PR, in prep for the sound driver changes - coming in the next PR.

If this is small, I am wondering what you consider big... :-)

> #### 1. I updated the linux make build to support optional ENV variables - to target a specific ROM images.
> This allows the targeted builds in much the same was the the windows command line build script supports.
> 
> The env variables are:
> 
> `ROM_PLATFORM`: - to speficy a ROM build - eg : RCZ80
> `ROM_CONFIG`: - to specify a config - defaults to std
> 
> If `ROM_PLATFORM` is not set, then it does as it previously did - and builds all images.

Sounds like an improvement.  May cause some confusion with builders that are used to putting a platform and config in the makefile directly.

> 
> #### 2. Update sub makefile references to use `$(MAKE)`
> There are uses of the following pattern:
> `cd <dir>; make`
> 
> I have changed them to use the make env pass thru:
> `$(MAKE) --directory <dir>`
> 
> This allows for some make switches to pass thru - eg -B. And also helps if parallel builds are activated - it be nice to get the Makefile to fully utilise parallel builds....
> 

Sounds good.  I sometimes find that minor Makefile changes cause edge case issues for some, so you may need to help me with those if reported.

What does ".ONESHELL" do?  Is it compatible with MacOS and Linux?

> #### 3. Some additional gitignores.

OK

> #### 4. Whitespaces!
> I am sometimes my own worst enemy - I have my development tools setup to adhere to some strict work patterns - its how I have encouraged the teams I work with.
> 
> By avoiding trailing spaces - it reduces noise in commits/merges.
> Tell me to stop it if you want - I dont wont to come across as trying to demand a specific and pedantic coding style.
> 

I think I would like working with you.  Probably hard to believe, but I do occasionally clean up whitespace in files as I am working on them.  I should think about automating this in my toolset.

> #### 5. Driver function table check - hard error
> I add a hard `FAIL` command within the preprocessor code that check the drivers have implemented the required count of functions. I found sometimes i had got it wrong on the sound driver - and never noticed the failure - now if the count is not as expected - the assembly will fail.
> 
> So
> 
> ```
> #IF (($ - XXX_FNTBL) != (XXX_FNCNT * 2))
> 	.ECHO	"*** INVALID XXX FUNCTION TABLE ***\n"
> #ENDIF
> ```
> 
> becomes
> 
> ```
> #IF (($ - XXX_FNTBL) != (XXX_FNCNT * 2))
> 	.ECHO	"*** INVALID XXX FUNCTION TABLE ***\n"
> 	FAIL
> #ENDIF
> ```

Great.

PR #113, opened 2020-04-25
Dean/dev 4

vipoo:
This PR primarly contains the changes to `tune.com` to use the new hbios sound driver.

It now has an optional switch `--hbios` which needs to be supplied after the filename

I believe without the switch - it operates as before - driving hardward directly.

It does not support `MYM` file type - and returns a message if u attempt to play `MYM` files with the --hbios switch

It does not support the noise channel yet.  I have been able to play the sample files: `ITERATN.PT3`, `SYNCH.PT3`, and `SANXION.PT3` - they seem to sound ok - not as good as the AY chip -- but you do notice the lack of noise channel support.

The resolutions of `periods` between the SN76489 (10bits)and AY chips(12 bits) does mean there is potential for some files to sound rather odd - if the value overflows --- one of the things to work on.

In updating the `tune.asm` code - i did extract a few bits of code into some seperate files - i just find it a little easier if associated code is grouped together.

In addition, I also have a updated to the TMS driver - to have an optional build flag to enable the sys timer from the VSYNC interrupt.  

> (I had also had a commit to modify the Ticks freq from the default of 50 - but i when i went to merge i saw you have done it.... - so the code i did should have been fully removed)

Cheers
Dean


wwarthen:
Looks good Dean.

I have a couple things I want to investigate, bit will merge and follow-up later.

Thanks,

Wayne

PR #123, opened 2020-05-09
Dean/dev 12

vipoo:
1. Updated sys time get function, to also return the tickfreq in register C.

2. Removed from documentation, description for a non-existing function.  It was something that i had added to aid my testing - its not required in the main build (i actually still have it as a patch file, that i apply when i am running specific tests)

3. minor fix to the debug tracing for the sn76489 sound driver.


wwarthen:
Thanks Dean.

PR #126, opened 2020-05-10
Dean/dev 14

vipoo:
Hi @wwarthen  and @b1ackmai1er 

I have a PR here that has refactored how the NOTE playing logic works.

It reduces the need for a full table map for every note - and only needs a mapping for one octave.

This has reduced the code size a bit, and make it easier to align the note values between drivers.

So calling NOTE for any driver with a value of $98 will always produce the note C4 (assuming the clocks of the sound chip are set correctly.)

I had some issues with my SN76489 chip. It appears that the doc for the SN76489 kit has an error in the Jumper settings - what i thought was CLK/4 was only CLK/2.  (Maybe its a revision difference...)

Anyway, once i had that sorted - i validated that the SN and AY cards are produceing the correct notes (maybe with a slight error of 1/2Hz to 1Hz or so)

I see there is some changes to the spk.asm ---- there  is a lot there i dont understand - I guess this is a single bit speaker?  I wonder how it would go if you used the Tune.com app ??  I would reckon a bit funny :smirk:

Cheers
Dean


vipoo:
I guess i need to update the architecture doc for the NOTE function, to indicate the mapping table values.

b1ackmai1er:
Nice work Dean. Funny you had the clk divide setting problem, I had the
similar issue with by ECB board at the beginning and it took me a long time
to figure out.

spk.asm is not compatible with Tune - it is only single channel and
requires cpu to constantly drive it. But still worthwhile including under
the hbios architect umbrella.

Will checkout you new code as soon as commit goes though.

Regards Phil.

wwarthen:
Looks good Dean.  Especially like the reduced note table size.

One note for you.  In some of the updated config files, I see that the terminal linefeed has been removed.  It turns out that TASM has issues with this.  It does not know to insert a  line feed at the end of an included file.  It frequently does not matter because the first line of the subsequent file is normally a comment, but I pulled my hair out a couple times with this issue.  I will add the terminal linefeeds, but just wanted you to understand.

Thanks!

Wayne

wwarthen:
> One note for you. In some of the updated config files, I see that the terminal linefeed has been removed. It turns out that TASM has issues with this. It does not know to insert a line feed at the end of an included file. It frequently does not matter because the first line of the subsequent file is normally a comment, but I pulled my hair out a couple times with this issue. I will add the terminal linefeeds, but just wanted you to understand.

OK, well guess it was actually only once cfg file.  :)

vipoo:
Hi Wayne,

Thanks for the heads-up - i didnt consciously delete the trailing newlines
- i have updated my editor settings to ensure there is always a trailing
newline in any file i edit.

Dean.

wwarthen:
No problem at all.  I learned this the hard way and just wanted others to learn from my struggles.  :-)

PR #127, opened 2020-05-12
Dean/dev 2

vipoo:
More sound stuff

1. Updated AY38910 driver to use systimer interrupt handler to manage the boot beep.  One of the cool things with retro systems - is the instant on.  I experimented with using the timer handler vector hook to setup a schedule function to turn off sound after a short period - instead of blocking during driver initialisation - it works ok on my machine.  I had to do some #define stuff to identify if there is a timer interrupt active.  If there is no timer interrupt assembled, then it just falls back to a blocking wait.  

Not sure your view on how i have done this, and I would not be surprised if  I have introduced issues i am not aware of.

2.  Tune - added a switch to adjust tone up or down octaves - this helps when playing back - if clocks are misaligned, or with the SN76489's smaller octave range.

3. updated the docs for the sound driver NOTE function, and description for Tune.com.  I am not great with the words :confused: - I am sure the descriptions could be made clearer/easier/simpler....

Cheers
Dean

FYI: @b1ackmai1er


wwarthen:
Thanks Dean.  Sorry for being a little slow with this merge, I'm a little overloaded.

I will want to review the timer interrupt mechanism for the startup beep, but that will happen when I have more time.

Thanks,

Wayne

wwarthen:
Hi Dean,

I am replacing the nice sharp and flat Unicode symbols you used in the Architecture document with plain old '#' and 'b'.  Sadly, the PDF processor is choking on them.  There is probably a way around it but it is messy because of the document generation flow I am using.  Sorry.

-Wayne

wwarthen:
Hi Dean,

Did a very brief bit of testing using my SC126 w/ AY sound card.  Everything generally worked well.  

I'm amazed that the HBIOS based interface plays back so well.  I couldn't tell the difference.  I'm pretty sure I did it right because if I disabled the HBIOS sound driver and used the --HBIOS startup flag, it said the hardware is missing.  When using the --HBIOS startup flag, the banner still says "timer mode".  I wonder if it could be changed to say "HBIOS mode" to confirm what is happening?

I do see a couple issues with the use of the system timer to shutoff the beep timer:

- The routine to shutoff the sound (AY_SETV) ultimately invokes a routine that disables and re enables interrupts.  The problem is that interrupts wind up re enabled inside of the interrupt handler.  It is essential that interrupts not be re enabled until the interrupt management framework does this on interrupt return.  This did cause my SC126 to crash until I removed it.

- The unhooking of the interrupt vector does not account for the possibility that something else may hook the vector between the time that you hook it and release it.  Properly unhooking a vector is messy.  You need to walk the chain until you find your position, then remove yourself from the chain.  The possibility of something else hooking the vector in this scenario is high because it is done during system initialization.

If you want to work on resolving all this, you can, but it feels like a lot of complexity to me.

Thanks,

Wayne

vipoo:
Hi Wayne,

Thanks for the review - appreciate the feedback details.

I am more than happy to work towards correcting the issues you have identified - kind of important i would think.

Not sure how long it will take me - do you want to revert the specific commit?

Dean.

b1ackmai1er:
Hi Dean,

In relation to the reset issue, I can confirm I get the same behaviour on
the ay without "slow down code" on my 10Mhz machine. i.e. device does not
play beep at boot up - just get a click. If I switch my slowdown flag on
(sbcv2004) which changes the cpu speed to 5Mhz during register writes, all
is good.

Note that this is just the Z80 CPU speed, not the AY clock.

This same code is required in TUNE to work as well and the usual behaviour
is that songs will play L+R channel but not center channel (noticable when
i play my favorite creep.pt3). This may be related as the ay code only does
the beep on the center channel. The reason I raise is that maybe with a
delay between channel writes, your 76489 may reset correctly.

Regards Phil.

PR #128, opened 2020-05-13
Dean/dev 3

vipoo:
Hi Wayne,

Here is my attempt to fix the interrupt handling issue i introduced in the AY driver.

It feels a little messy.

I understand how i broke it - but i am not able to confirm the fix - as I never reproduced the issue you had.  Can u advise on configuration that would expose the bugs?

Its was tricking this.  I have done the following:
1. During interrupt handler, I suppress the DI/EI operations.
2. I return an error state for sound functions until the boot beep has completed.
3. I no longer remove then handler, I just simply no-op the handler.  I am not sure how to 'walk' the link list - i can know where the next handler is - but i can't see where it's next handler is?

I had thought about what you mentioned before, regarding moving the boot beep into a higher order logic somewhere  -- an optimisation I would like to look at, as time permits

One set of code I did not submit - was some code i worked on to allow the drivers to play notes for a duration - and que up other notes to play immediately after.  It was working ok.  I wanted this for the code changes I was doing in bbcbasic - as it has this requirement - but i ended up moving the logic into the bbcbasic code base itself.

The bbcbasic has to 'poll' the timer, to work out duration periods - where as the code I had in HBIOS was triggered by the timer interrupt.  

It didnt feel right to have this logic in a BIOS layer - its a higher level feature - it cant go in the OS - so i put it basic. I guess for their time, the basic interpreters were their own kind of OS.

I'd be interested in your thoughts about what kind of behaviour should and shouldnt be in HBIOS?

Cheers
Dean.


wwarthen:
Hi Dean,

Thanks for this.  I'm trying to work myself out of a code disaster at the moment.  I hope yo don't mind if I sit on this for a day or two while I get things put back together.

Regarding the failure scenario.  I was running an SC126.  The build was the standard SCZ180_126 build, but I just enabled the AY driver.  That was the only config change I made.  My system would start playing the beep, but when the beep ended, the system usually froze.  Seems like almost any Z180 would do the same thing.

I agree that adding a full blown playback buffer inside of the BIOS seems like too much.  The idea of the BIOS is to abstract the hardware.  I think playing an audio buffer is more than the BIOS should do.  Playing a series of notes seems like an application level thing and BBC BASIC can be considered an application in this sense.

Thanks,

Wayne

wwarthen:
Hi Dean,

OK, I am going to go ahead and merge this.  I did not review the code yet, but easier to review after merging.

I forgot to mention that I updated the cfg_state.inc file.  It needed more parens in the conditionals.  Another quirk of TASM is that is does not do proper prioritization of conditional operators -- just goes strictly left to right.  When I initially tested SC126, I realized that the new interrupt based beep code was not used.  Turns out that it was because of the stupid TASM order of evaluation thing.  I have learned to use liberal parens in conditionals with TASM!

It sounds like your changes to the interrupt-based beep should work.  You are correct that it is actually not possible to walk the interrupt handler chain.  I was not thinking correctly.  That means you cannot remove yourself from the chain.  Your solution of staying in the chain, but no-op your handler is all you can do.  I'm not thrilled with leaving an entry in the chain, but OK for now.  I will think about a way to set single use timers in HBIOS that clean up after themselves, but that will be later.

I will post back after testing this merge.

Thanks,

Wayne

wwarthen:
Based on cursory testing, the changes seem to be working.  At least, the previous failure scenario is not recurring.

Thanks!

Wayne

vipoo:
Nice!

And thanks again regarding another TASM insight

b1ackmai1er:
Hi Wayne,


The device ids for the sound boards is 01h, 02h etc. Other device
enumeration in hbios are10h, 20h etc.

Should we change these to match existing convention?

i.e

SND_SN76489 = 0x10
SND_AY38910  = 0x20

Thanks.

wwarthen:
> Hi Wayne, The device ids for the sound boards is 01h, 02h etc. Other device enumeration in hbios are10h, 20h etc. Should we change these to match existing convention? i.e SND_SN76489 = 0x10 SND_AY38910 = 0x20 Thanks.

That's a really good question.  Some history... the device id's were originally coded with the idea that the high nibble would be the device type and the low nibble would be the relative unit number within the device space.  When I overhauled the API to maintain a blob of configuration data with each unit number, this was no longer necessary.  Since there were places in the code that depended on the device ids, I did not change them.  However, my plan is to rationalize them back to be just sequential numbers.  So, by happenstance, the device ids for the sound drivers are actually fine and in line with my long term intentions.  :-)

Thanks,

Wayne

b1ackmai1er:
cool. thanks

PR #129, opened 2020-05-19
Dean/dev 4

vipoo:
Hi Wayne,

I have some enhancements to HBIOS's heap management - added code to allow for deleting blocks.

I wanted to be able to access that memory space, with dealloc support.

I had to change the header - to include a back pointer - to make heap walking a little easier.  Not sure if that has any unforeseen consequences? 

I have tested it on my unit, and run it against some automated tests.

I see there is an enhanced error code returned for failure.  I added a macro to help - unless you object, Ii might have a go at applying that throughout.  Has an added benefit of getting rid of a call to empty function with disabled.

Totally understand - If you want to reject, for any reason.

Cheers
Dean


wwarthen:
Hi Dean,

I do have some questions on this before merging this.

I had started to do exactly what you did, but never wound up committing it.  I never found a use case for it because there is no way to reorganize used/free space.  It does add complication and I would like to understand the use case.  What do you intend to use it for?  One more question.  I don't understand how HEAPCURB is handled.  When HB_RESET is called, it is critical that all memory allocations subsequent to HEAPCURB be released.  I am also unclear how the deallocated blocks are used.  It doesn't look like your code is searching the free blocks at allocation (maybe I missed this, I was scanning quickly).

On the CALL_SYSCHK macro, I actually started to do that myself, but backed off.  There are two problems: 1) it does not handle a common scenario where you want to do something like: "CALL NZ,SYSCHK" and 2) I want to make the SYSCHK display dynamically configurable in the future.  #2 is the most important.

On the RET_SUCCESS and RET_FAIL, I have a concern similar to #1 above.  They cannot be used in all cases.  I struggle with macros that cannot be used consistently.

Let me know about the memory dealloc use case and we will go from there.

Thanks!

Wayne

vipoo:
Thats cool.

There are 2 things here - i should have separated them in the PR.

1.  I will double check the HEAPCURB - and make sure its reset correctly.  I assume the new RESET (COLD/WARM) can be use to for testing??

I wanted my applications to be able to load code into the HBIOS free space (there is a fair amount sitting there idle) - and then invoke it.  Idea 2 in Issue #131 .  I need my app to be able to alloc space, use it, then free it before exit - otherwise memory is not released.

I tried to limit the amount of code in the garbage collector - so it does not realloc holes in memory - only release memory at end of heap - but this was all thats need for app usage -- and there is single app usage only - the risk of memory holes is small.

2.  In regards to `SYSCHK`.  (I appreciate that the macro system in TASM is, well, how can i say this nicely - very limited.  As i have been learning z80 assembly, i find that there are a **lots** of assemblers available - and they all seem rather limited in someway or other!)

This was my thinking, but totally understand your concerns.

For me, replacing this code that appears about 2 dozen times in the code:
```
	CALL	SYSCHK
	LD	A,ERR_NOFUNC
	OR	A
	RET
```

with this:
```
RET_ERR(NOFUNC)
```

greatly aids readability, ensure consistency, and avoid duplication, and enable easy tweaking in future.

With the addition of:

```
#IF (SYSCHKENABLE)
#DEFINE CALL_SYSCHK	CALL SYSCHK
#ELSE
#DEFINE CALL_SYSCHK
#ENDIF
```
We can avoid a call to a empty function, when the check is disabled - saving a few precious bytes.

If need be, it could be inlined, when the macro is not adequate to the task.

Just me thoughts.

vipoo:
HEAPCURB

I haven't, I believe, change the behaviour of this field.  This still track the memory allocated by HBIOS.  OS and app allocation happen after. 

If i do a `RESWARM` reset, all my application memory is released - as i would expect.

If a do a `RESINT`, all os and application memory allocations are released

Is there something I am missing?

Cheers
Dean

wwarthen:
Thanks Dean.  See below...

> 1. I will double check the HEAPCURB - and make sure its reset correctly.  I assume the new RESET (COLD/WARM) can be use to for testing??

You can use the warm start to test, but not cold start.  Cold start blows everything away, so it is not a valid test.  You will need to set a breakpoint and confirm the integrity of the allocation chain after allocation and deallocation scenarios.  Looked a bit more at the code and I guess the HEAPCURB handling is OK.

> I wanted my applications to be able to load code into the HBIOS free space (there is a fair amount sitting there idle) - and then invoke it. Idea 2 in Issue #131 . I need my app to be able to alloc space, use it, then free it before exit - otherwise memory is not released.

There are some serious issues with this use case.

1. An application cannot expect memory to be available in the HBIOS bank.  This varies widely depending on configuration.  Some existing configurations leave almost no space.

2. If an application aborts, memory will not be freed.  Under CP/M, such aborts are commonplace and not under the control of the app.  Any disk I/O error prompts the user to abort.

3. Any application that messes with HBIOS memory bank can easily and accidentally corrupt HBIOS.  The Z280 allows absolute protection of such a memory space and I like the idea of doing that which would preclude application usage.

I had intended that memory allocations in HBIOS be used by OSes to allocate disk buffers because having such buffers in that bank is a major performance improvement in certain cases.  For applications, it makes much more sense to me that the application would want to use a separate RAM bank.  Much more memory and much less chance of corrupting HBIOS.  Also, such usage could be implicitly freed when the app exists cleanly or aborts.

I see how this feature is conceptually good, but am struggling with the practical reality of the situation.  Given the fact that an application cannot count on any memory available in the HBIOS bank, what kind of application would benefit from this?

I'm not trying to stand on principle here, I just want to be sure that any added code and complexity within HBIOS is warranted.  I am going to want a very cogent example of how this will be used before moving forward with it.  Idea #2 in #131 is really specific enough.  I'm trying to visualize a specific application and scenario that would benefit.

> I tried to limit the amount of code in the garbage collector - so it does not realloc holes in memory - only release memory at end of heap - but this was all thats need for app usage -- and there is single app usage only - the risk of memory holes is small.

OK, got it.  However, I'm not sure that is much more useful than the simpler model of allowing an application to free from a given "point" in the allocation.

> 1. In regards to `SYSCHK`.  (I appreciate that the macro system in TASM is, well, how can i say this nicely - very limited.  As i have been learning z80 assembly, i find that there are a **lots** of assemblers available - and they all seem rather limited in someway or other!)
> 
> This was my thinking, but totally understand your concerns.
> 
> For me, replacing this code that appears about 2 dozen times in the code:
> 
> ```
> 	CALL	SYSCHK
> 	LD	A,ERR_NOFUNC
> 	OR	A
> 	RET
> ```
> 
> with this:
> 
> ```
> RET_ERR(NOFUNC)
> ```

I guess I am OK with this.  I have mixed feelings such macros.  Maybe a personal problem, but I tend to miss implications in code when the code is inside a macro.  It definitely cannot be used in all cases, but I suppose it is OK for the majority as you point out.

> With the addition of:
> 
> ```
> #IF (SYSCHKENABLE)
> #DEFINE CALL_SYSCHK	CALL SYSCHK
> #ELSE
> #DEFINE CALL_SYSCHK
> #ENDIF
> ```

This is my main issue.  I absolutely do not want to avoid the call to the empty procedure.  The reason is that I want to be able to enhance the SYSCHK procedure in the future to dynamically decide whether to display the error or not.  That requires that procedure always be called.

In summary, I am OK with RET_SUCCESS and RET_FAIL.  I am also OK with RET_ERR(n).  I am not OK with CALL_SYSCHK because I always want the SYSCHK procedure called for future enhancement.  I will add that I am a little reluctant about RET_FAIL because it makes it very easy to return the generic error code which should really not be used going forward.

Sorry for all the back and forth.  I do appreciate all of your efforts!

-Wayne

vipoo:
Cool - hope I am not being annoying with the back forth -- appreciate your time.

> In summary, I am OK with `RET_SUCCESS` and `RET_FAIL`. I am also OK with `RET_ERR(n)`. 
Cool

> I will add that I am a little reluctant about `RET_FAIL` because it makes it very easy to return the generic error code which should really not be used going forward.
I agree - i created `RET_FAIL`, before i saw your new stuff - i would agree we drop this and only use `RET_ERR`.

> I am not OK with `CALL_SYSCHK` because I always want the `SYSCHK` procedure called for future enhancement.
Thats cool.  Maybe if you have time, help me understand what your goal is here.  I assume its something only activated when debugging, diagnosing issues?  On a production build it would not be active at all? I feel like i am missing something here.

Cheers
Dean

vipoo:
I have created too many topics - I am losing track of all the items I raised.  Hmm.

In regards to using HBIOS heap - like a lot of this stuff - its me just hacking away - not always understanding - so appreciate your feedback.

I will close this PR - and follow up with the other points in discussion.

Thanks
Dean

wwarthen:
All fine by me.  I enjoy the exchange.

For now, my perspective is that I want to get a much clearer idea of the need, then we can revisit all potential options.

It is probably easier to close this for now, but may revisit depending on the application needs you come up with.

Thanks,

Wayne

wwarthen:
> > I am not OK with `CALL_SYSCHK` because I always want the `SYSCHK` procedure called for future enhancement.
> > Thats cool.  Maybe if you have time, help me understand what your goal is here.  I assume its something only activated when debugging, diagnosing issues?  On a production build it would not be active at all? I feel like i am missing something here.

Of course.  Sorry I wasn't specific before.  The idea is that the RomWBW HBIOS Diagnostic Level (just a number) is stored in a byte in the HCB.  It defaults to a build-time setting, but it can be updated as desired.  Initially, I am going to add a function to `romldr.asm` to allow a user to modify the value before booting.

When a SYSCHK is encountered, control will go to the SYSCHK routine just like now.  However, instead of a build-time conditional, the routine will check the Diagnostic Level value.  If it is over some pre-defined value, SYSCHK will display the diagnostic message.  If not, it will return.

On a production build, the default value for the Diagnostic Level would be either 0 or something fairly low, meaning only display critical messages.  Of course, the level could be changed at will.

Ultimately, this Diagnostic Level would control all of the HBIOS messaging.  Well, not the INIT stuff, but all of the messages the drivers can spit out during regular processing, especially disk I/O errors.

I have not fully thought through the specific level numbers nor have I thought through the implementation in the drivers.  SYSCHK is really just a small piece of the puzzle, but it is a good and useful start.

I would be thrilled to hear thoughts from all concerned.  But please don't code anything related to this without checking with me first.

Thanks!

Wayne

wwarthen:
I just committed a very preliminary version of this.  The diagnostic level can be set in the loader.  SYSCHK will display if the diagnostic level >= 8.  Diagnostic levels are defined in std.asm.

vipoo:
Not look at your code yet - but it sounds like a log level system -
something similar to many logging libraries (FATAL, CRITICAL, WARN... INFO
etc) - or the Syslog Message Severities idea.

Cool

wwarthen:
Nowhere near that sophisticated, but that is the general concept.

PR #133, opened 2020-05-22
DUART driver

codorjan:
Serial driver for 2681/88C681/26C92/etc.


wwarthen:
Looks pretty clean Chris.  Will merge now and then do confirm all of my standard builds work.  I will typically have some questions after I do that and will post them here.

Thanks!

Wayne

wwarthen:
I just did a follow-up check-in to allow other configs to build cleanly.  Otherwise, looks good.  I have no way to test currently, so will forward any usage issues to you.

Is there any reason that the DUART driver configuration cannot be folded into the standard RCZ80 config?  It appears to have all the auto-detect logic as needed.  Collapsing it into the main config would be more consistent and easier for users I think.

Thanks!

Wayne

codorjan:
Thanks!

I agree that it should eventually be placed in the standard configuration, although I was hoping it could first be tested with other devices that might be decoded at 0xA0 (i.e. ACIA and UART). I don't have RC2014 modules for either of those at the moment, but I _do_ have an SC115 breakout and the relevant ICs so I can wire those up (hopefully over the next few days) and see for myself how things go.

wwarthen:
Sounds good Chris.  I will leave the RCZ80_duart configuration out there while you are testing.

Thanks,

Wayne

PR #138, opened 2020-06-05
DUART: fix detection of SC26C92 devices

codorjan:
I got ahold of some Philips SC26C92 DUARTs and... they don't work. Turns out a reserved bit is always set to 1 when register MR0 is read and it was causing them to be misdetected. (Exar DUARTs read the bit as 0, so it worked as expected on the XR88C92.)

Fixed by masking out all but bit 1, since we really only care if it's set (i.e. MR0 exists) or not.


wwarthen:
Thank you!

Issue #137, opened 2020-06-03
Retrieving hardware configuration information


vipoo:
Hi Wayne,

Hope all is well.

I have been away from RomWBW for a bit (the usual work demands)

I am trying to catchup with where we at with the ability to retrieve hardware configuration stuff.  I am not sure i remember which approach you decieded to go with.

I would like to be able to access some hardware attributes.  I had a look at the code and docs.  And observed the following:

a. The TMS ports.  I see code for the `BF_VDAQRY` function, returing the `TMS_DATREG`.   It only returns one port - I guess the assumption other ports are assume to be incremented from this.

b. The SN76489 has a new function `SN7_DEVICE` - which returns the `SN76489_PORT_LEFT`.  It does not return the right port number.  

c. The AY driver has a similiar configuration. 

d.  The sound drivers also have a sub-function `BF_SNDQUERY`, that allows for the retriving of all io ports.

e. I see a new functions (`VDAFN`, `SNDFN` and similiar others for other devices)
These seem to return an address in DE for driver specific config.  The  Sound drivers are returning 0 and the TMS driver is returing the `TMS_PPIx` entries - but nothing else.

I may not have interpreted this right, and it may also be still under-development, so my questions below may be jumping the gun a bit:

1.  Is the expectation that all drivers will return key attributes through the `xxxFN` functions?  With the other access methods deprecated?

2.  Should the TMS driver's data table also include `TMS_DATREG` and `TMS_CMDREG`

3.  The TMS driver also has a mirror of the internal registers `TMS_INIT9918`.  It be nice to be able to read this also, so that i can determine if the interrupt line has been enabled or not.  To avoid disabling that.  (I will also raise another seperate issue - for a challenge i have in regards to using the TMS and interrupts)

3. Can the Sound drivers implement this fully to not only return the IO ports, but also clock frequency?

I am happy to help do some of the coding here - if this is what you want.

I have been playing with my fork of bbcbasic, and got it doing plotting lines and pixels through the TMS chip.  It be nice to able to retrieve the IO ports from HBIOS, instead of having it hard-coded in the assembly.  It be also nice to get information that would enable me to inject the right level of wait states for IO access.  

The bbcbasic fork also uses the sound driver interface to enable full sound capability.

Cheers
Dean

vipoo:
Cool,

Sounds like I have something to work on this weekend - its a long weekend here - would you believe my country celebrate a Queens Birthdays this weekend? 

[...]

PR #140, opened 2020-06-06
TMS: update the VDAFN data payload to include ports and register mirror

vipoo:
Hi Wayne,

TMS Commit for issue #137, for your review

If you happy with this, i'll work on something similiar for the sound chips.

Cheers
Dean

### DIFF
--- a/Source/Doc/Architecture.md
+++ b/Source/Doc/Architecture.md
@@ -428,7 +428,7 @@ _Code_ | _Meaning_
 -7     | media not present
 -8     | hardware not present
 -9     | I/O error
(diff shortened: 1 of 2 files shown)


vipoo:
Was not sure how you might have wanted to document the payload structure for each driver, so i added an Appendix section to the doc.

wwarthen:
Looks great Dean.  Thank you.

I like the appendix concept.  In truth, the appendix probably ought to be somthing that describes each of the drivers along with the driver's configuration data and options.  Unfortunately, I don't have the energy to put into that right now!  :-)

PR #145, opened 2020-07-15
Add port configuration for new Rev 6 AY sound card

electrified:
Tested on Z180, no reason to believe it shouldn't work in the Z80 build too.

Both OS driver and tune work.

Open to ideas about a better name for the constant than `AYMODE_RC_REV6`. Do you think `AYMODE_RC_MSX` would be better? Or even just `AYMODE_MSX`?


wwarthen:
Hi Ed,

Changes look perfect.  I can't test right now, but see no reason to worry about them.  Will merge momentarily.

I like using AYMODE_MSX since the config would apply to MSX port config generically.

Thanks,

Wayne

PR #149, opened 2020-07-28
Support for Tiny Z80 SBC

skiselev:
This patch adds support for [Tiny Z80](https://github.com/skiselev/tiny_z80) SBC
The board is supported as **tz80** configuration for Easy Z80 platform.


wwarthen:
Thank you Sergey!

PR #150, opened 2020-08-01
Z80 EIPC Support - Add and use register definitions

skiselev:
This patch adds definitions for Z80 EIPC (Z84C15) registers and their values. It also uses these definitions in the initialization code.


wwarthen:
Thanks Sergey!

PR #154, opened 2020-10-10
EIPC_MEM_xWS values corrected

mlukasek:
There were duplicities in EIPC.INC, so it was not possible to build EZZ80_tz80.rom


wwarthen:
Thank you Martin.  Somehow this was already fixed in the dev branch, but not the master branch.

-Wayne

PR #159, opened 2020-12-04
Update to flashfs

b1ackmai1er:
First working code allowing a 4k sector in low ram to be written to flash


wwarthen:
Thanks and congratulations on getting this done Phil.  I am really backlogged at home with non-hobby work, but I will try to play with this in the near future.

wrljet:
What is this thing "non-hobby work"?  I'm not familiar with that term.  :-)

b1ackmai1er:
Hi Wayne, would appreciate your thoughts on the driver implementation - when you get a chance.

Current behaviour - boot defaults to ram and rom drive using md driver.
Future behaviour -

1. New flash driver initializes before md driver.
2. If it detects 1 or 2 39SF040 chips it adds the flash memory as a DIO using flash filesystem driver (to be implemented), sets a flag for flash filesytem enabled.
3. md driver initializes. If flash filesystem flag enabled, DIO entry for adding ROM disk is skipped.

New disk type ID would be required i.e. 0xA0 Flash Memory Disk and probably a new hard disk drive attribute i.e. Bits 5-3 : Type 7 - ROMFlash

OR 
Incorporate the flash filesystem into the md driver.
Flash disk remains as Disk device type 0x00 Memory disk.
If it detects 1 or 2 39SF040 chips it adds the ROM disk DIO as normal but with hard disk attribute Type 7 - Romflash
md driver checks MD_CFGTBL attribute entry to determine handling within the md driver.

So essentially the question is should the r/w flash be implemented as a seperate device driver type or incorporated as a Memory disk subtype.

Regards Phil

wwarthen:
> So essentially the question is should the r/w flash be implemented as a seperate device driver type or incorporated as a Memory disk subtype.

My gut instinct is to integrate Romflash and md.  Although I have not spend any significant time with the Romflash code, I have to think yoiu have already done all the hard work.  I envision a config setting that enables "write capability" in a single driver.  Since Romflash requires a 4K RAM buffer in HBIOS address space, I think it is important to allow this to be enabled/disabled.

A variant on my preference would be to simply enhance the Romflash driver to be a full replacement driver for md.  Romflash would need to be enhanced to do handle RAM drives, but I think that should be pretty straightforward.  In this scenario, the user could simply choose which driver they want.

Please don't take my preferences as anything more than my opinion -- you asked.  :-)

Thanks again for this contribution!

-Wayne

PR #163, opened 2020-12-15
Beta release of flash file system driver

b1ackmai1er:
There are still bugs and lots or room for code improvements but probably worthy of a beta release

Enable with

MDFFENABLE	.SET	TRUE		; MD: ENABLE FLASH FILE SYSTEM 

Shouldn't impact normal operation with it disabled. Maybe make sure you have a spare flash chip :)

Regards Phil


wwarthen:
I briefly reviewed things Phil and I'm comfortable with the changes when MDFFENABLE is FALSE, so it shouldn't cause any pain to existing users.  I will test and spend more time on the actual FF code as soon as I can.  I am going to merge now.

Thanks!

PR #164, opened 2020-12-16
Flash file system optimizations

b1ackmai1er:
Inline code, optimizations


wwarthen:
Nice optimization Phil.

Thanks.

wwarthen:
Hi Phil,

Did a little testing of the FF code today.  Working quite well!

The only thing I noticed (and you probably already planned to handle this) is that you do not trap the use of flash chips other than the 39F040 and disable flash writes.  For example, I iniitally had a 29F040 chip in my system and that just fails to write.

Thanks,

Wayne

b1ackmai1er:
hi wayne, ah yes i am only checking for 39sfo40s. with verbose output
enabled does it say Flash filesytem enabled?

does 29fo40 support 4k sectors. i dont think it does.

will check tonight.

wwarthen:
The 29F040 has a sector size of 64K, so it is not a good candidate for the Flash Filesystem.  I guess I was just thinking there should be more of a warning in the boot messages that the chip discovered is not supported.  At first, I didn't even realize I had a 29F040 in my system.

With debug enabled, I see:

MD: FLASH=1 1= 00000000 0000 0000 A401 0000  ID:A401 29F040
MD: UNITS=2 ROMDISK=384KB RAMDISK=256KB

None of this is a big deal.  Virtually everyone is using the 39F040 chips.  Not sure how I wound up with a 29F040.

-Wayne

b1ackmai1er:
hi wayne. there is a debug equate and a verbose equate.


with verbose set it should show chips identified. if it is a 39sf040 and
md_ffenable is set then it should say "flash file system enabled" but
doesn't say anthing if those conditions are not met.. i can add "flash
filesysem unavailable" if you prefer.

wwarthen:
> hi wayne. there is a debug equate and a verbose equate. with verbose set it should show chips identified. if it is a 39sf040 and md_ffenable is set then it should say "flash file system enabled" but doesn't say anthing if those conditions are not met.. i can add "flash filesysem unavailable" if you prefer.

Hi Phil,

Sorry, realized I never got back to you on this.

Yes, you are correct, with verbose mode enabled (which it is by default), then the chip type is displayed.

I see what you are saying that "FLASH FILE SYSTEM ENABLED" only appears if the chip detected is one that is supported by the flash file system.  I also see that the code disables the flash file system if a 39F040 is not detected.  That is good.  I misunderstood what I was seeing previously.  The code is doing the right stuff.

With respect to the boot message, I would suggest that you print either FLASH FILESYSTEM ENABLED or FLASH FILESYSTEM DISABLED depending on whether a supported flash chip is detected.  These messages would only be displayed if MDFFENABLE is set to TRUE in the config.  So, basically, if the user has requested FF in his/her config, they the user will get an explicit message indicating if it is in use or not.  This assures the user they they have properly enabled the config regardless of whether the right chip is available.

Thanks,

Wayne

b1ackmai1er:
hi wayne, thanks for the feedback. will update. Have you had any failures
reading or writing?

PR #167, opened 2020-12-23
Add verify after write as default, restructure local stacks in flash file system

b1ackmai1er:


wwarthen:
This looks good Phil.

I like the way you improved the local stack management.

I see a couple nits that I will correct after merging.

Thanks!

Wayne

PR #168, opened 2020-12-24
md.asm minor optimizations for flash file system

b1ackmai1er:


Source/HBIOS/md.asm
```diff
@@ -999,36 +996,34 @@ MD_S_SZ	.EQU	$-MD_FERAS_R		; SIZE OF RELOCATABLE CODE BUFFER REQUIRED
 ;          IX POINTS TO DATA TO BE WRITTEN
 ;          A  CONTAINS CURRENT BANK 
 ; ON EXIT  NO STATUS RETURNED
+;          AF' TRASHED
```
wwarthen, line 999: This may be an issue for some application software.  Some CP/M application software assumes that the OS/BIOS will not touch the alternate register set.

wwarthen:
I like the changes Phil.  Makes the invocation of the high memory functions easier to understand.

I do have a small concern about trashing ZF'.  There is some CP/M application software out there that assumes the alternate register set will not be modified by the OS/BIOS.  In general, HBIOS therefore stays away from use of the alternate register set.  It is OK to use it as long as you restore anything modified before returning to the calling application.

I am going to merge this and let you consider the AF' usage.

Thanks,

Wayne

b1ackmai1er:
Thanks Wayne, I will look at that.

Regards Phil.

PR #169, opened 2020-12-25
Update md.asm

b1ackmai1er:
Further optimization of loops utilizing alternate register AF'


wwarthen:
Thanks Phil!  Looks good.

PR #175, opened 2020-12-30
Support System Update process for systems with 39SF040 flash chips 

b1ackmai1er:
Generate an update file.
Updated docs.
md.asm minors
Update compression application
Compressed update file not supported yet.


wwarthen:
Hi Phil,

This all looks good, but would you mind if I hold off on merging this for a few days?  I am away from home this week and I would like to be able to check out the build changes on my development system.

Also, note that I will be moving your changes to ReadMe.md to to Source\Doc\GettingStarted.md.  This is because ReadMe.md is generated from Source\Doc\GettingStarted.md, so your changes would get wiped out when I regenerate the doc files.

Thanks,

Wayne

wwarthen:
Hi Phil,

OK, I am about to merge this.  Good stuff.

In Build.ps1, I am wondering why you used a loop to concatenate files to create the .upd file instead of just doing it in one line like the .rom and .com files?

I will do some testing once I complete the merge and make sure all the builds are working.

Thanks,

Wayne

wwarthen:
Hi Phil,

It looks like the Windows version of lzsa has been updated in this check-in.  As a result the compressed font files seem to be different between Windows and Linux builds.  Can you look into this?

Thanks,

Wayne

b1ackmai1er:
hi wayne, can change it to concat. Initially thought it  would be better in
case you wanted to modify but that does not really make sense in retrospect.

regards phil

b1ackmai1er:
Didnt think of that. Both should decompress the same. Will take a look.
Might need to ask the gent who did the linux build to assist. There is also
no problem with reverting this but will try and sync the two.

Regards Phil

b1ackmai1er:
oh i remember was planning on compressing each bank and then appending them
one at a time using the same temp file.

transferring, decompressing and then flashing seemed to hard and i chicken
out :)

b1ackmai1er:
hi wayne iv added the liinux version but it is not working so will need to
do more work on it before it can be merged.

regards phil

wwarthen:
Thanks for pursuing this Phil.  I think it is important to keep the different builds binary identical.  Otherwise, it will be make testing much more time intensive.

-Wayne

PR #180, opened 2021-01-12
updater bank select fix

b1ackmai1er:
Added crc32 calculation.
Still outstanding - issue if transmission is not started within 5 seconds.


wwarthen:
Thanks Phil.

I know I still owe you some thoughts on how to make updater.asm available more easily than someone building a custom ROM.

PR #182, opened 2021-01-15
Extract flash routines from md.asm and updater.asm to common library flashlib.inc

b1ackmai1er:
This completed the af/af' and push/pop cleanup.
Calling and relocating routines not included as md and updater use different methods 
Checked out ok on sbc and mkiv


wwarthen:
Excellent.  Thanks Phil.

PR #183, opened 2021-01-17
Updater and md improvements

b1ackmai1er:
md - Flash relocating routine now only copies  what is required rather than full 64 bytes every time. Move code space check into flashlib

flashlib - Add chip erase support. Consolidate code space check here. Only display on error.

updater - Add support for other 512K flash devices. Devices that do not support sector erase will be chip erased prior to writing. Only AM29F040 tested. Additional debug options added


wwarthen:
Looks good Phil.  Thank you.

PR #185, opened 2021-01-22
Adjustments for new hbios proxy location

b1ackmai1er:


wwarthen:
Thanks Phil.  I will merge these changes.

The 128 byte expansion of the proxy should be temporary.  Getting the Z280 native memory support working has been challenging and I had to make some adjustments just to get started.  I'm making good progress and (in theory) I should soon be checking in an update that relocates this extra 128 bytes out of the proxy area.

Regardless, the changes you made are very good because they make the code automatically adapt to the start of the proxy.  In a perfect world, all code should automatically adapt to the start of the proxy at build time.  I made some similar changes in other code as I discovered dependencies on the proxy start address.

-Wayne

wwarthen:
@b1ackmai1er 

I just checked in another round of Z280 updates.  The expansion of the proxy space has now been reverted back to it's original size.

-Wayne

PR #191, opened 2021-02-09
acia - RTS fix and minor updates

feilipu:
This is a fix for the RTS configuration, to make the setting specific.

Also
 - fixed whitespace, change tab to 8 spaces for consistency.
 - makes the acia reset usage consistent and correct.
 - small optimisation in checking for received bytes.

There are some other files touched, but there should be no changes except whitespace (trouble with git that I can't fix, ugh...).
That's all I need to touch. :+1:


wwarthen:
Thanks Phillip.  All looks fine to me.  Will merge now.

PR #219, opened 2021-07-20
Update RAM fail routine

lynchaj:
Change RAM fail to abort on error in page and move on to next page


wwarthen:
Thank you Andrew!

PR #226, opened 2021-08-01
DMA fall-back mode for md.asm driver

b1ackmai1er:
By default, if dma is not present or fails to initialize, software will be used.


wwarthen:
Thanks Phil.  Good update.

-Wayne

PR #227, opened 2021-08-14
Round reported CPU Mhz up if  within 20hz of next MHz increment

b1ackmai1er:
Example: 9.995Mhz will now be reported as 10Mhz


wwarthen:
Thanks Phil.  Good idea.  I am merging this now.

Does it make sense to round up if within 500KHz of next MHz increment?  In other words, true rounding.

-Wayne

b1ackmai1er:
Hi Wayne,

If it is not going to impact any timing routines then it is the logical
think to do I think.

Best Wishes, Phil

wwarthen:
> Hi Wayne, If it is not going to impact any timing routines then it is the logical think to do I think. Best Wishes, Phil

I implemented the true rounding.  Should be best in all cases.  I like your code for doing this by the way.  Very slick.

-Wayne

PR #234, opened 2021-08-24
zmp support command line port setting 

b1ackmai1er:
Refer notes.txt


wwarthen:
Thanks Phil.  I like your simple solution to setting the com port.  That should have occurred to me.  Thanks!

PR #238, opened 2021-08-28
dbgmon enhancements

b1ackmai1er:
(p)rogram command will now display value of current byte.
r(un) command will now return to the monitor if program executed a return. 
On return registers are displayed. 
Run can now optionally pass values to A and BC


wwarthen:
Excellent work Phil.  I am a bit worried about space.  dbgmon is tight on space when the ROM is built with DSKY enabled.  I am going to merge your changes and check it out.

Thanks!

b1ackmai1er:
thanks wayne. can tabs be used to format the help instructions?

wwarthen:
Looks like space is OK.  I am still seeing 100 bytes slack space when building with DSKY enabled.  FYI, DSKY mode NG uses the most space.

wwarthen:
> thanks wayne. can tabs be used to format the help instructions?

Hmmm... I've always stayed away from that because it is hard to say what different terminals will do.  With that said, the only situation I can think of where a terminal would not do 8 position tabs is if someone intentionally programmed their VT100 with custom tabs.  Probably safe to use tabs.

wwarthen:
BTW, I actually had a very small dbgmon change pending on my end.  You will see it in my next check in.  I noticed that the dump command was displaying characters with the high (8th) bit set.  That can cause all kinds of mess on some terminals.  So, I added an additional filter so that those values are substituted with dots just like control chars.

b1ackmai1er:
Thanks Wayne. Note that last update overwrote romldr and rom applications
update.

Best wishes Phil

wwarthen:
> Note that last update overwrote romldr and rom applications update. Best wishes Phil

That does not sound good.  I will try to fix that later today.

PR #242, opened 2021-09-07
Forth fixups, improvement and documentation updates

b1ackmai1er:
Would be good to allocate more rom space for this on the next memory shuffle - 512b for customer forth words


wwarthen:
Thank you Phil!

PR #243, opened 2021-09-10
Add ability to Transfer a file to memory by x-modem

b1ackmai1er:
HI Wayne, Is this something you would be interested in? Have not extensively tested but seems to work. The main consideration is the extra space it uses. let me know if you want to proceed and I will continue testing and optimizing.


wwarthen:
I don't have strong feelings about it because the most common way to transfer files to a monitor is via simple intel hex upload.  However, I generally like the capability.  The problem is definitely the space usage.  If DSKY is enabled, this would put us way over the available space for the monitor.

Thanks,

Wayne

b1ackmai1er:
should i wrap it up in a conditional that defaults to not include it.

maybe put it in boot loader instead

cheers phil

wwarthen:
> should i wrap it up in a conditional that defaults to not include it.
>
> maybe put it in boot loader instead cheers
>
> phil

I like wrapping it in a conditional -- that would be fine for now.  I don't think it would be very useful in the boot loader and the boot loader is also pushing the limits of size.

Thanks,

Wayne

b1ackmai1er:
Thanks wayne will do.

b1ackmai1er:
Hi Wayne, cheated a bit - x-modem option will only enable when dsky is not :) hope that compromise is ok

wwarthen:
Thanks Phil.  I'm merging this PR now.

-Wayne

PR #244, opened 2021-09-13
Doc updates

b1ackmai1er:


wwarthen:
Excellent, thanks Phil.  I will merge and then run the document regeneration script and then push that.

-Wayne

PR #247, opened 2021-09-22
Doc updates and preparation for run time memory resizing

b1ackmai1er:
Removal of some "hard coded" ram size calculations
Dynamically update the bank codes in the HCB
MBC bank selection code change needs to be verified.
Other processor banking may need adjustment too.
There is probably a lot wrong with this but though it would be a good starting point.
Of coarse the actual memory size testing code is missing. The assumption is that this runs and populates 
CB_RAMBANKS and from there the bank ids are adjusted.


wwarthen:
Wow, sounds great Phil.  Due to the nature of these changes, I am going to need to review very carefully.   Please give me some time to do that before I merge this.

Thanks!

Wayne

wwarthen:
Hi Phil,

This is really nice work.  I especially like the way you reorganized the RAM/ROM bank definitions in std.asm.  Although these definitions become a little misleading when RAM size is dynamically determined, I still think they are useful to document the memory bank layout.  Ultimately, some of these equates may simply need to become comments referring to the HCB fields.

I did see a couple things we should discuss before I merge. 

First, in the BNKSEL routine for MBC, you have added some code to dynamically handle the required XOR of a bit depending on memory size.  This is the right idea, but won't work for 128K and 512K systems.  For those systems, **no** bits are XORed.  Also, there is a pretty significant performance penalty for this added code.  I think the best approach is to use self modifying code where there is a single line for the XOR: "XOR %00000000".  As is, the line will do effectively nothing.  During boot initialization, if the memory size is either 256K or 1024K, you would dynamically modify the XOR operand to be either %00000100 or %00010000 respectively.  What do you think?

Second, in the SPK driver, you have added code to round the CPUKHZ value when deriving the best possible CPUMHZ value.  This is a good idea, but the CPUMHZ value is already rounded when the CPU speed is dynamically calculated.  I notice that I have **not** done the CPUMHZ rounding when the CPU speed is not dynamically calculated.  I would think the best approach is to add the rounding to the CPUMHZ value that is calculated in std.asm.  What do you think on this one?

Thanks,

Wayne

b1ackmai1er:
Hi Wayne,

Thank you. I have submitted an update for the mbc bank selection code. I
definately made some wrong assumptions there :)

Could you take a look and test or provide feedback?

Regarding the speaker code, the change is to actually remove the rounding
that was in place that I thought was not required now.
Yes I think we should round the the CPUMHZ. It did try it but could not
stop it overflowing and generating incorrectly.

Regards Phil

wwarthen:
Thanks Phil.

My mistake on the CPUMHZ thing.  I read the diff backwards!!!

I will take a look at the revised MBC code and merge if it all looks OK.  Yes, the MBC bank mapping thing is really confusing.  I will try to add some more comments.

Thanks,

Wayne

wwarthen:
OK, as far as I can tell from looking at the code, this all looks fine.  I will merge and then do some testing.

Thanks Phil.

b1ackmai1er:
I expect that this has a fair bit of evolution  to do :)

wwarthen:
> I expect that this has a fair bit of evolution to do :)

Definitely.  However, you have made a lot of progress.

PR #250, opened 2021-10-01
Update md.asm - Table size off by one. Fix endien confusion for flash id 

b1ackmai1er:
Table size off by one. Fix endien confusion for flash id byte


wwarthen:
Thanks Phil.

PR #253, opened 2021-10-18
Add new video function

b1ackmai1er:
For your consideration. Seems to work on the ecb-vdu. 

### Function 0x4F -- Read a character at current video position (VDARDC)

| _Entry Parameters_
|       B: 0x4F
|       C: Video Device Unit ID

| _Exit Results_
|       A: Status (0=OK, else error)
|       E: Character
|       B: Character Color Code
|       C: Character Attribute Code

Read a character from the current cursor position including it's colour
and attributes. If the display does not support colours or attributes
then return colour white on black and no attributes. If the device does
not support the ability to read a character, return error status

### DIFF
--- a/Source/HBIOS/vdu.asm
+++ b/Source/HBIOS/vdu.asm
@@ -265,6 +265,10 @@ VDU_VDASCR1:
 ;----------------------------------------------------------------------
 ;
 VDU_VDARDC:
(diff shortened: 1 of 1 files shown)


wwarthen:
Makes sense to me.  Code updates all looked fine.  Need to implement the function in the other video drivers someday, but I'm going to merge this as is.

Thanks,

Wayne

PR #254, opened 2021-10-18
Implement TastyBasic as a RomWBW submodule

dimitrit:
This PR aims to resolve two issues:

- it fixes an issue  (dimitrit/tastybasic#1) raised by @b1ackmai1er 
- it implements TastyBasic as a git submodule (cf. https://github.blog/2016-02-01-working-with-submodules/), making it easier to keep TastyBasic in sync between the two projects

I appreciate that adding submodules has an impact on existing code maintainers in that they need to explicitly fetch the tastybasic code when changes are made in the tastybasic repository. However, as the tastybasic code is intended to be (relatively) stable meaning code changes should be infrequent, the impact should be manageable.


wwarthen:
Thank you Dimitri.

I am generally fine with this concept.  I have not used submodules before, so give me a little time to understand this, then I will merge.

Wayne

dimitrit:
Hi Wayne

I've just discovered that this version of Tasty Basic does not work in simh.  As I assume it should work without having to make simh specific modifications, I suggest we don't merge this PR until I've figured out what's wrong.

Many thanks,
Dimitri

PR #270, opened 2021-10-30
Update TastyBasic

dimitrit:
This PR replaces tastybasic with the latest version:

* previously 'missing' statements are implemented - DATA,READ,IN,OUT
* a CP/M version (`TBASIC.COM`) is included in the CP/M 2.2 and ZSDos utilities
* the CP/M version includes SAVE and LOAD statements allowing programs to be persisted on disk


wwarthen:
Thank you Dimitri.  I am tied up most of the day today, but will merge these changes as soon as possible.

wwarthen:
Hi @dimitrit 

OK, just finished working though everything.  You did a lot of work here.  Thank you!

I am going to merge this now.  You will see a check-in from me shortly after.  This is just to adjust some build stuff to get a binary consistent build across all platforms -- no functional changes.

Note that I did need to remove the automatic imbedding of the git version in tastybasic.  This is because it causes the build to fail if not run from within a git source tree.  Many people just download the source tree as a zip file and build it.  In that scenario the git commands fail.

Thank you,

Wayne

PR #272, opened 2021-11-01
Include TastyBasic version details

dimitrit:
Hi Wayne
Thank you for merging my TastyBasic PR (#270).   

This PR reinstates the version details in the Makefile, which will allow me to keep track of the TastyBasic version included with RomWBW. Note that VER is now a hardcoded value, which corresponds to the [release tag in the TastyBasic repository](https://github.com/dimitrit/tastybasic/releases/tag/v0.3.0).

Thanks
Dimitri


b1ackmai1er:
Hi Dimitri,

Thanks for doing the update :)

wwarthen:
This works nicely.  Thank you Dimitri.

-Wayne

wwarthen:
Hi @dimitrit 

Sorry, I think I am going to need to revert the change you made that deletes the .rom output files when using "make clean".  Doing a make clean is not intended to remove the final outputs, only the files built in the Source directory.  Only make clobber should do that.

Thanks,

Wayne

dimitrit:
Hi @wwarthen 
Without wanting to start a symantic discussion about what `make clean` should or should not remove, I included this change for two reasons:
* I have been caught out too many times chasing down 'old' issues because the rom images were not being rebuilt after making changes to source files. This suggests an issue with the make dependency graph, which might take a while to chase down. Since I habitually run `make clean` before doing a build removing all rom images in clean was a 'quick and dirty' fix.
* The `Clean.cmd` script in the Binary directory removes rom images and does not appear to support 'clobber'

Apologies if I inadvertantly broke things.
Dimitri

wwarthen:
Hi @dimitrit,

No need to apologize.  The situation is a little messy and I can see why it was frustrating for you.  I didn't actually put together the make build system, I just try to keep it working.  For many years, there was no way to build under Linux, so I was very happy that something was contributed and worked.

The Windows build system came long before the make build system.  In the Windows build system, you run "Clean.cmd" at the level of the tree that you want cleaned.  In a distribution build, Clean.cmd is run from the Source directory to leave the contents of the Binary directory in place.

The make build system was constructed differently.  It uses "make clean" to clean everything but the Binary directory.  It is equivalent to running Clean.cmd from the Source directory.  "make clobber" is equivalent to running Clean.cmd from the root.  In a distribution build, make clean is used to leave the contents of the Binary directory in place.

This inconsistency has existed for some time and I guess I just got used to it.  I will look at making these consistent in the future.

Thanks again for your contributions.  The updated TastyBasic seems to be working great (although I don't use it very deeply).

-Wayne

PR #278, opened 2021-11-27
VGMPLAY Updates

b1ackmai1er:
Assume default file name VGM, basic memory protection for large files


wwarthen:
Thanks Phil!

I'm not home, so cannot test, but this looks straightforward.

PR #307, opened 2022-11-26
ym2612 stub and hbios driver loading code cosmetics

b1ackmai1er:


b1ackmai1er (commented):
Hi Wayne, sorry I've included some changes in this that break build. Working on update

b1ackmai1er:
OK I think I have fixed the build failure. The Driver debug standardization is to enable simpler enabling of debugging output via the custom cfg file rather than in the driver itself 
Example DEBUG_DRV .SET DRV_ID_PCF8584 will enable debug output in that driver (if it exists) and can be further combined with your TRACE work.

wwarthen:
Thanks for all of this Phil.  A lot of files have changed, so I am going to need some time to work through all of this.  I may have some questions.

-Wayne

wwarthen:
Hi @b1ackmai1er,

Sorry for the delay.  I finally got some time to look through everything.

Generally looks fine and is building without errors for me (haven't tried Linux/MacOS yet, but should also be fine).

Lots of grammar fixes -- thank you!

The YM2612 stub is fine.

On the changes to the DEBUG build management, I definitely agree with the idea of getting all of these DEBUG conditional config variables into the config files where they belong.  I did run into a conundrum with the use of a single variable to control all of them though.  I don't see a way to enable multiple DEBUG sections at the same time.  This is something I do quite a bit.  Also, what about DEBUG sections in the main HBIOS or other shared code areas?  I do see that the single config variable reduces the number of config variables from many to one.  Thoughts on this?

Thanks,

Wayne

b1ackmai1er:
Hi Wayne,

Yes I realized this limited the debugging to one driver but thought this
was a reasonable decision. However if you have more than one active at a
time then am happy to back this out since you do all the development and
debugging work and use it a lot.

Do you have any ideas? I did consider using a bit map/mask but there are
too many.

I was planning on going through the hbios and also see if i could
define/group some debug modules too, like interrups, z180,z280, memory
mapping.

Regards Phil

b1ackmai1er:
I have an idea for enabling two drivers for debugging and will try it out.

wwarthen:
Hmmm... I am struggling with the idea that the number is limited at all.  Allowing 2 is better, but not sure that really works for me either in the end.  Sorry if I am being a pain.

Thanks,

Wayne

b1ackmai1er:
no problem, I'll back out the changes this evening.

Regard Phil

wwarthen:
I sincerely apologize for this Phil.  I really hate to see work backed out.  I support the idea that the DEBUG controls should be in the config files, but I don't want to restrict the usage to 1 (or a few) being turned on at a time.  I really appreciate all of your efforts!

Thanks,

Wayne

b1ackmai1er:
Not a problem at all.

Regards Phil

PR #313, opened 2022-12-31
Change romldr to accept a baud rate

b1ackmai1er:
Accept a baud rate instead of a code


wwarthen:
Thanks Phil.

The baud rate changing interface is a very nice improvement.

-Wayne

PR #317, opened 2023-02-19
vgmplay updates, romldr baud improvements, hbios/cbios low memory handling and other minors

b1ackmai1er:


wwarthen:
Thanks Phil.  There is a fair amount of change here, so I will need to work through everything.  I might tag the individual changes with comments in GitHub.

-Wayne

wwarthen:
Turns out that I am comfortable with all the changes.  Some very nice work here Phil.  Thank you!

Let me know if you have any other changes brewing in the near-term.  I will want to coordinate merge timing because I am starting the final round of regression testing for a new stable release.

-Wayje

wwarthen:
> Just going to try and tidy up that last disk type string table. Other than that, I don't have anything significant in the wings and am happy to hold off until the new release.

I am fine with tidying stuff up.  That is what I am doing now as well.  I would like to hold off on anything more significant though (for now).

Thanks,

Wayne

PR #322, opened 2023-03-06
DMA updates

b1ackmai1er:


wwarthen:
Thanks Phil.  I'm not really testing the DMA code -- just making sure it does not break the build.

-Wayne

b1ackmai1er:
These were just cosmetic, so should be no change to behaviour (crosses
fingers behind back while standing on one leg)

wwarthen:
I did make a single character change to overcome a build issue on Linux.

PR #330, opened 2023-03-18
Update dma.asm

b1ackmai1er:
cp/m fails to boot with the dma driver included but no dma hardware installed. This is because the default "dma_fail_flag" indicates that the dma is good to go. But in safe mode the dma initialization never takes place to find no dma and set this flag, so the memory driver hangs. This change is to make the default status of the flag the failed state so the memory driver will fallback to software in recovery mode.


wwarthen:
Nice catch @b1ackmai1er.  Thanks.

PR #336, opened 2023-03-22
Fix default ZETA V2 SBC configuration

elevendroids:
- Set the default CPUOSC to 8MHz (default value in the project's documentation), updated the User Guide
- Set the FDD controller to ZETA2 - regression introduced in bd664c3


wwarthen:
Thank you @elevendroids!

-Wayne

PR #344, opened 2023-04-25
HARDWARE TIMER AND PS2 KBD INTERRUPT SUPPORT

b1ackmai1er:
The PS2 keyboard driver uses the same codebase as the current kbd.asm driver so I have set it up so that it will not assemble when the other drivers using kbd.asm are active. All experimental. Will work on further integration.


b1ackmai1er:
Need to rethink integration

PR #351, opened 2023-07-01
Add ColecoVision-compatible configuration

jblang:
Adds `RCZ80_jbl.asm` configuration with ColecoVision-compatible settings:

```
; THIS CONFIGURATION ENABLES DRIVERS FOR THE TMS9918 AND SN76489 BOARDS BY J.B. LANGSTON
; AND THE YM2149 BOARD BY ED BRINDLEY. THE TMS9918 IS CONFIGURED TO USE THE COLECOVISION
; PORTS AND HAS INTERRUPTS DISABLED BECAUSE COLECOVISION USES NMI, WHICH WOULD BREAK CP/M
; WHEN ENABLED. IT IS CONFIGURED FOR A 3.686MHZ CPU CLOCK IN ORDER TO BE COMPATIBLE WITH
; COLECOVISION GAMES. THE FIRST SIO PORT IS CONFIGURED TO RUN AT 115200 BPS WITH THE CPU
; RUNNING AT THIS SPEED. THE SECOND SIO PORT IS CONFIGURED TO RUN AT 115200 BPS WITH THE
; SECOND CLOCK SET TO 7.373MHZ. LOWER BAUD RATES CAN BE ACHIEVED ON THE SECOND PORT BY
; MOVING THE SECONDARY CLOCK DIVIDER JUMPER (E.G., 38400 @ 2.458MHZ). THE BAUD RATE DOES
; NOT NEED TO BE CHANGED IN THIS CONFIGURATION FILE IN ORDER TO DO THIS, BUT THE DEVICE
; LIST WILL INCORRECTLY SHOW THE PORT RUNNING AT 115200 REGARDLESS OF THE CLOCK DIVIDER.
; UNCOMMENT THE LINE THAT SETS BOOTCON TO 1 TO BOOT ON THE SECOND SIO PORT BY DEFAULT.
```


wwarthen:
Thank you J.B.  Very nice addition to RomWBW and glad that it is working.  I am merging this now.

How would you feel if I changed the configuration file name to RCZ80_col (short for ColecoVision)?  I certainly want to ensure you get proper credit for this and would add a credit to the file comments.  It is not a big deal to me, so happy to leave it as is if you prefer.

Thanks!

Wayne

jblang:
It wouldn't have mattered to me, but you've already merged it.  Anyway I thought all the cool kids put their initials in things :).

PR #355, opened 2023-08-20
Update DMAMON

b1ackmai1er:
Add missing option to toggle ready latch. Default to Duodyne configuration.


wwarthen:
Thanks @b1ackmai1er!

Shortly, I will bump the build version number and publish a new Development Snapshot.

Thanks,

Wayne

PR #356, opened 2023-08-21
Duo/MBC IM2 Hardware Timer support

b1ackmai1er:
This is some work to recognize the DUO and MBCs IM2 pin header interrupts.
And specifically to support a hardware time interrupt using them.
Further work to be merged to support a VT82C42 based keyboard using IM2 pin header arrangement with a circular buffer.
Interrupt allocation/numbering is unclear as there is no real standard and limited inturrupt slots means overlaps.


wwarthen:
Thanks @b1ackmai1er!

PR #357, opened 2023-08-23
Fix references to the EPFDC driver in comments.

fernandocarolo:
The EtchedPixels floppy disk controller is known by the identifier 'FDMODE_EPFDC', but the comments in the Config subdirectory have an incorrect reference to 'EPWDC'. This fix updates the comments to use the correct name.


wwarthen:
Thank you @fernandocarolo!

PR #359, opened 2023-09-16
Dev

danwerner21:
This should correct a few issues with the ESP32 driver.   All of the lines that I changed should be tagged with my initials.
These changes have been tested on the Duodyne, but should not have any issues on the Nhyodyne.

The only file modified should be esp.asm, however a "make clean" was executed prior to commit.


wwarthen:
Hi @danwerner21,

Thanks for this.  Changes to esp.asm look great.  The PR has 33 files changed though.  I don't really want to take all of that and then try to clean it up.

I can just extract the esp.asm changes and check that in if you want.  Alternatively, if you want to try and get the PR down to just esp.asm, that would be great.

Let me know what you think.

Thanks,

Wayne

danwerner21:
I am not sure what is going on with the other files as they all appear to be binary adds.   Is there a different cleanup script I should be running other than “make clean”?

 

I started this effort with a clean clone of your dev branch, and only changed the esp.asm.  (although I did temporarily create a custom DUO configuration, but ended up deleting it).

 

Dan

wwarthen:
Hmmm... `make clean` should have done it.  Did you run it from the high-level directory (RomWBW itself)?

danwerner21:
yup -- I am going to try this again.   I need to figure out why I cannot get you a clean PR.

Issue #362, opened 2023-09-23
RPi4: Build failure "zxcc: Cannot locate bios.bin" when path depth is less than 4 levels


skullandbones:
@wwarthen I attempted to build the dev branch with HEAD commit 092e44c62e450a21a37809ce842981b33d8dc317

I decided to do a test build in the /run directory which is a RAM based tmpfs file system.

The path was /run/test/RomWBW, I did `make` but the build failed:
```
SLR180 Copyright (C) 1985-86 by SLR Systems Rel. 1.32 #NL0029

 zmo-rw01/HF
End of file Pass 1
End of file Pass 2
 0 Error(s) Detected.
 937 Absolute Bytes. 91 Symbols Detected.


../../../Tools/Linux/zxcc MLOAD25 -ZMP.COM=ZMPX.COM,ZMO-RW01
+ ../../../Tools/Linux/zxcc MLOAD25 -ZMP.COM=ZMPX.COM,ZMO-RW01
zxcc: Cannot locate bios.bin

                            make[3]: *** [Makefile:10: zmp.com] Error 1
make[3]: Leaving directory '/run/test/RomWBW/Source/Apps/ZMP'
make[2]: *** [../../Tools/Makefile.inc:136: all] Error 2
make[2]: Leaving directory '/run/test/RomWBW/Source/Apps'
make[1]: *** [../Tools/Makefile.inc:136: all] Error 2
make[1]: Leaving directory '/run/test/RomWBW/Source'
make: *** [Makefile:3: all] Error 2
```
I did some experiments by using a fresh git clone of RomWBW in various directories to see what worked and what failed:
/tmp/RomWBW - build failed
/tmp/temp/RomWBW - build failed
/tmp/temp/temp2/RomWBW - build succeeded

So it seems that there is a problem building bios.bin as it appears to depend on the depth of the path to the RomWBW directory.

I didn't see this failure while looking at issue 358 because my path was /home/pi/projects/romwbw

Dean

skullandbones:
In the good case:
```
/run/test/temp/temp2/RomWBW$ find . | grep bios.bin
./Tools/Linux/bios.bin
./Tools/zxcc/bios.bin - is in git repo
./Tools/unix/zxcc/bios.bin
./Tools/unix/zxcc/cpm/bios.bin - is in git repo
./Source/pSys/bios.bin
```

In the bad case:
```
/tmp/RomWBW$ find . | grep bios.bin
./Tools/zxcc/bios.bin
[...]

PR #363, opened 2023-09-24
zxcc: Fix for load bios.bin failure on a RPi4

skullandbones:
This is a resolution for Issue #362

Fix load_bios() when used on a Raspberry Pi4.

The defect was corrupting the path of the zxcc executable.

The fix is to NUL terminate the path string from the readlink() call.


skullandbones:
I removed the redundant codeblock as the local len variable could be defined in the existing codeblock of the if statement..

wwarthen:
Looks great @skullandbones.  Thank you.

Merging now...

-Wayne

PR #367, opened 2023-10-08
Uploading HTalk program to supplement talk.com program

TomPlano:
Uploading HTalk program to supplement talk.com program. Similar functionality, but uses HBIOS calls and Char IDs, as opposed to CP/M calls and IDs. Ive also annotated optional code and regions of code copied from dbgmon.asm. Where possible I used the I/O routines from dbgmon.asm (including exact label names) which should allow easy integration into the monitor, space permitting.

### DIFF
--- a/Source/Apps/HTalk/Build.cmd
+++ b/Source/Apps/HTalk/Build.cmd
@@ -1,0 +1,14 @@
+@echo off
+setlocal
+
+set TOOLS=../../../Tools
+set PATH=%TOOLS%\tasm32;%PATH%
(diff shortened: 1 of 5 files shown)


wwarthen:
Thanks @TomPlano,

It looks like there is no impact to the build process, so I am going to merge now.  I do see that another line needs to be added to the Windows Build.cmd file in the Source/Apps folder, but I will just add that with some other stuff I am about to check-in after your stuff.

Thanks,

Wayne

PR #372, opened 2023-11-17
Allow override of secondary console front panel switch selection

dylanhall:
As discussed in the retro-comp forum.
This patch defines the configuration setting SECCON which allows overriding the default increment behaviour when using the PRI/SEC front panel switch.


wwarthen (approved):

wwarthen:
Looks good.  Thanks @dylanhall.

PR #383, opened 2024-01-07
Typo in Source/ReadMe.txt

feilipu:
Just a typo.

You added that just to check we actually read it, right? ;-)


wwarthen:
> Just a typo.

Thanks @feilipu!

> You added that just to check we actually read it, right? ;-)

Sure, let's go with that...  😀

PR #393, opened 2024-04-16
NABU Personal Computer integration - nabu.asm and nabukb.asm contain …

lesbird:
…code to support the NABU hardware.

PR contains code and modified code to support the NABU personal computer.

NABU.ASM is optional and included for future updates to support an interrupt timer that will also call in to HB_TICK to update the RomWBW timer. At the moment with the current hardware this does not work and the interrupt never gets called.

NABUKB.ASM is the code to support the NABU keyboard. The NABU keyboard has a watchdog timer that sends $80 every 3 or so seconds. The only way to filter this out is to read the data port and check for the watchdog timer byte and replace it with a $FF before returning to the RomWBW keyboard code. This drove me crazy as I had no idea why I was getting random bytes from the keyboard. After reading the NABU documentation I discovered the issue and was able to work around it.


Source/HBIOS/nabukb.asm
```diff
@@ -0,0 +1,86 @@
+;======================================================================
+;	NABU KEYBOARD DRIVER
+;
+;	CREATED BY: LES BIRD
+;
+;======================================================================
+;
+; KBPORT EQU	$90
+;
+; POLL FOR INPUT
```
lesbird, line 10: This commented out code example is from the NABU ROM which I used as a reference

Source/HBIOS/nabukb.asm
```diff
@@ -0,0 +1,86 @@
...
+	XOR	A
+	RET
+;
+NABUKB_READ:
+	IN	A,(NABUKB_DAT+1)
+	AND	$02
+	JR	Z,NABUKB_READ
+	IN	A,(NABUKB_DAT)
+	LD	E,A
+	CP	$80
```
lesbird, line 71: If we receive the watchdog timer byte we replace it with $FF and it is ignored by RomWBW and the operating systems.

wwarthen:
Thanks @lesbird.  I'll get this merged tomorrow.

-Wayne

wwarthen:
Thanks again @lesbird.  All of your changes look fine and I am about to merge them.  There are a few things that I would like to tweak just for improved consistency with existing RomWBW stuff.  Hope that is OK.

Thanks!

Wayne

PR #409, opened 2024-08-22
Adding in Heathkit H8 support

lesbird:
Adds features to the RomWBW code base to support the Heathkit H8 computer with my RomWBW CPU board. The Heathkit H8 has a very unique front panel display and keypad. The front panel generates interrupts at 500Hz at which time it is expected to be serviced. Servicing the front panel requires the following steps:

1. It is mandatory to read port 0xF0 of the front panel to clear the pending interrupt
2. Each seven-segment LED of the front panel, of which there are 9, will need to be refreshed - one LED per interrupt. LEDs turn off after a number of milliseconds (hardware design) when not updated which is why it is important to refresh them during the interrupt cycle. When they are refreshed the off-time in hardware is reset.
3. The horn (speaker) bit is updated to determine if it should emit a sound or not - this is bit 7 when writing to port 0xF0
4. A 2ms timer is updated and stored at location $000B - most Heath CP/M software uses this timer for software timing
5. For RomWBW I call into HB_TICK every 10 interrupts (50 times/sec)
6. For RomWBW I added a "system up and running" timer (clock showing hours,mins,secs since powered on) to the front panel
7. The numeric keypad is sampled to determine if any special functions should be performed - IN from any I/O port, OUT to any I/O port, view a memory location, monitor a specific Z80 register (AF,BC,DE,HL,PC) - this is hacky, increase/decrease the CPU speed.

Updating the LED display is handled by first selecting the LED position (low 4 bits) combined with the speaker bit (7) and writing that to port 0xF0 followed immediately by the LED pattern value and writing that to port 0xF1. Pattern values need to be inverted when writing to port 0xF1. Writing a value is 0 will turn on all segments and writing a pattern of $FF will turn off all segments.

Note that the default numeric display mode for the Heathkit H8 computer is split octal notation. Where the top of a 64K memory location in hex is FF.FF in split octal it is displayed as 377.377. Years later Heathkit began to adopt HEX in some of its CPU upgrades with reprogrammed ROM chips but not many H8s had that upgrade. The front panel code included in this commit (H8P.ASM) uses split octal to maintain consistency with the original H8 system.

Reading the numeric keypad is simply a matter of reading port 0xF0 where a value is returned based on what key is depressed. If no keys are pressed then the return value is $FF. The keypad values are documented on my Wiki page here: https://github.com/sebhc/sebhc/wiki/H8-FP-(Front-Panel)-V1.x#low-level-front-panel-access

I had to steal some memory in the upper memory region (HBX_INTSTKSIZ) to store values to be exchanged between "app" space and "interrupt service" space. Within this stolen memory lies 3 bytes of data and about 14 bytes of code. The 3 data bytes are used to store a memory location (2 bytes) and the value stored at that location (1 byte). This data is used when the front panel is in memory viewing mode where a location in "app" space memory can be viewed on the front panel display. The associated code is used to copy this data to high memory so it can be accessed during the interrupt servicing routine. Additionally this code also copies the 2ms tick counter from high memory down to "app" space low memory at $000B. There's not a lot of available memory in high memory area so I tried to keep this to a minimum.

Saving Z80 register values is a bit tricky. It appears that BC,DE,HL were preserved so I was able to copy those registers as soon as the interrupt service routine was entered. The PC and AF registers were extracted by reading their values from the stack pointer. In H8 land the PC (program counter) is the location in memory that was saved as a return address when the interrupt happened - so the idea is that it shows the "current app space" code location in realtime.

Wayne, you might want to skip adding the BuildMSX.cmd script. That is up to you. It builds a CF image with CPM 2.2, ZSDOS, CP/M 3 and combines it with your hd1k_msxroms1 and 2 disk images. This will be the standard disk image that I will be using for my RomWBW CPU board for Heathkit computers moving forward.


wwarthen:
This all looks great Les.  I am going to retain the BuildMSX.cmd script, but it will not be executed as part of the official build.  So, it doesn't really cost anything, but will be easy to use.

Thank you for this contribution!

Wayne

PR #416, opened 2024-09-04
Genesis Modules initial commit

drj113:
Genesis modules STD card initial commit.

Support for GM STD Bus Z180 with onboard SD card driver
Support for GM IDE interface with Compact Flash drive


wwarthen:
Thank you @drj113.

The changes look fine.  I am proceeding with the merge.  There are a few nits that I will handle in a subsequent commit:

- ide.asm still has conditionals
- Remove hdsk.asm from HDIAG directory (I don't think you intended to put that there).
- Minor formatting (tabs vs. spaces in a couple places)
- GMZ180 is only built automatically under Linux/MacOS.  I need the Windows build to be consistent.

Thanks, Wayne

PR #421, opened 2024-09-07
User Guide FAT

kiwisincebirth:
These should be final changes to userguide. Merged FAT section into trnasferring Files, and other minor improvements


wwarthen:
Thank you @kiwisincebirth!

PR #426, opened 2024-09-19
gm7303 changes to support DSKY operations

drj113:
Changes to support GM7303 as a DSKY device

These changes firm up keyboard support in the GM7303 LCD / Keyboard module as well as the associated boot modifications in preparation to the dbgmon changes.


wwarthen:
Hi @drj113,

Thanks.  I am going to merge this PR shortly.  Just a couple notes:

- The assembler does not have an "if else" conditional construct (sadly).  I will be checking in fixes for that shortly after the merge.
- Just a reminder that I am working on the DSKY API, so you will probably need to make some changes related to that when I check-in that work.

Thanks, Wayne

drj113:
Hi Wayne,

Thanks for that.  It's very much appreciated.

I wonder what's different with my build environment.  The #IF ELSE
construct works beautifully here.

That's a bit frightening actually,  as it suggests something is different.

Doug

wwarthen:
First of all, I said "#IF ELSE".  That was a typo.  I meant to refer to "#ELSE IF".

I think you are using Linux or MacOS, right?  Under Linux/MacOS, the uz80as assembler is used.  Under Windows, the TASM assembler is used.  I know that the "#ELSE IF" construct failed to work as expected with TASM.  I didn't test it under uz80as, so it may work there.

Although uz80as is designed to be fully TASM equivalent, there are some subtle problems.  This is unfortunate.  Because of this, I have incorporated a process that confirms the TASM and uz80as builds are binary identical, so I am able to catch this type of thing quickly.

Did you see the construct "#ELSE IF" documented somewhere?  From what I see of both the TASM and uz80as documentation, that construct is not supported.  Perhaps this is an undocumented feature of uz80as.  If so, I sure wish TASM supported it.

Thanks, Wayne

drj113:
Ahhh.

I am a windows user :-)    I will start the dive now to see what assembler
is in my world.  I wonder if it is uz80as?  That seems to ring a bell
somewhere.  If I need to uninstall it, then so be it - the last thing I
need is subtle problems.

Thanks for doing the check in - I have been finding that if I keep working
on stuff, and not submitting it, when I do go to grab the most recent
branch, my whole world explodes  (I think I went from 80 to 84 most
recently) ...  (I am clearly *bad* at using these tools) - I have simply
duplicated directories and stored archival versions for my whole life.....

Kindest regards,

Doug Jackson

em: ***@***.***
ph: 0414 986878

Follow my amateur radio adventures at vk1zdj.net

wwarthen:
OK, so if you are on Windows, then you are using TASM.  I went back and looked and I think I understand what happened.  I think TASM was treating "#ELSE IF" simply as "#ELSE" and ignoring the rest.  It looks like this would have resulted in correct output for GMZ180, but it was causing issues for the other platforms which were using the "else" portion of the conditional.

I'm not quite sure why things are getting out of sync for you with GitHub.  However, it is generally easier to check-in frequently and not let your local work get too far out of date with the repository.  I don't mind the Pull Requests as long as I can confirm they do not affect the non-GMZ180 platforms.

Thanks, Wayne

drj113:
Ahhhhhh!!!

Thanks for checking that, and yes, I am definitely using TASM.

As I have mentioned before, the last thing I would ever want was to
break others work - I will make sure I don't use #ELSE IF in the
future.

Kindest regards,

Doug Jackson

em: ***@***.***
ph: 0414 986878

Follow my amateur radio adventures at vk1zdj.net

PR #428, opened 2024-09-20
Dean ez80 5

dinoboards:
# EZ80 Enhancements

1. Implemented ez80 specific DELAY and VDELAY functions (delegated to firmware) - ensure good consistent delay periods regardless of CPU frequency and W/S
2. Chang the default frequency config to 20Mhz - probably want the kits will come with as default
3. fixed some bugs/issues with the calculations for bus wait states.
4. Ported the sn76489 driver

Slight issue with the SN76489 (and I don't know if stock Z80 has this issue).  When the 2 stereo outputs are sent and combined to a mono speaker - the small phasing difference between the channels, can cause apparent cancellation of some sounds/notes.  Its dependent on the operating CPU frequency as to the effect.  When listening with separate speakers - not really a problem.


wwarthen:
Thank @dinoboards -- looks fine.

PR #429, opened 2024-09-22
ez80: updated reference to renamed firmware repo

dinoboards:
I have renamed the repo containing the ez80 'firmware' - probably not good etiquette to have it named 'rc2014-...'

This PR is just the updated comment


wwarthen:
Thanks Dean.

Issue #427, opened 2024-09-20
CBIOS Issue with High slice numbers in hd512


kiwisincebirth:
Just found an issue (assume it might be known) with assigning high slice numbers in hd512. 

SD0 (disk2) - is 8GB media with a 512MB RomWBW partition (hd1k)

```
Capacity of disk 2:  ( 15G)  31116288      Geom dacc1010
Nr  ---Type- A --      Start         End   LBA start  LBA count  Size
 1    RomWBW   2e      8:0:1  1023:15:16        2048    1048576  512M
 2 FAT32 lba   0c   1023:0:1  1023:15:16     1050624   15564800    7G
```

SD1 (disk3) - is 8BG media with FAT 384mb FAT partition (hd512) - this is actually imaged from a hd512_combo.img file

```
Capacity of disk 3:  ( 15G)  31116288      Geom dacc1010
Nr  ---Type- A --      Start         End   LBA start  LBA count  Size
 1             00       *** empty ***
 2     FAT16   06   1023:0:1  1023:15:16     1064960     786432  384M
```

So when I assign and try and use these

```
D>assign l:=sd0:255

   L:=SD0:255

   901 Disk Buffer Bytes Free

D>l:
ZSDOS error on L: No Drive
Call: 14

D>assign l:=sd1:255

   L:=SD1:255

   901 Disk Buffer Bytes Free

D>l:   << This is an issue since the drive is logged? without error >>
L>d:
```

... CBIOS correctly generates an error with hd1k Slice 255, since this exceeds the space in the $2E partition, 

But for but for SD1 (hd512) no such issue is generated, it just switches to the L: drive without error, any writes to the L: drive have potential to overwrite a valid partition.

I believe this wouldn't be hard to fix. For Hd512 - We just need to do the size comparison with the LBA start of the FAT partition (or any partition), instead of the total media size in CBIOS as it does now.

A second issue is the ASSIGN.COM is not doing any validation, which would probably require the duplication of logic in 2 places. I am not a fan of duplication, it is probably better to get the logic right in one place (CBIOS) first.

wwarthen:
Hi @kiwisincebirth,

> But for but for SD1 (hd512) no such issue is generated, it just switches to the L: drive without error, any writes to the L: drive have potential to overwrite a valid partition.

Yes, under hd512 there is no partition table entry that defines the size of the space to use for RomWBW.

[...]

PR #430, opened 2024-10-06
Added new HBIOS Function SYSGET_DIOMED

kiwisincebirth:
An enhancement to add a new HBIOS function SYSGET_DIOMED
Relates to issue #427


kiwisincebirth:
Note:  I am not sure why the readme files have such a large DIFF, i literally only updated the date, although I think this was maybe done by automation (?) which had other changes, which I ignored / reverted.

Note2: See the attached MS Word document which covers the testing of this function. If there is another test you want me to perform let me know ? If you want this file committed let me know.

[Results of Testing HBIOS Function.docx](https://github.com/user-attachments/files/17267824/Results.of.Testing.HBIOS.Function.docx)

wwarthen:
Hi @kiwisincebirth,

This looks great.

The readme files have a large diff because they are generated by Pandoc and the line endings produced are different when run from *nix than from Windows.  Not a problem.

The testing document is excellent.  Thank you for the thorough testing and putting it together.  I'm not sure there is a good palce for it in the main repository, but since it is attached to this issue, it should remain available for reference if anyone is researching this change.

I will merge this change momentarily.  The only additional thing I want to check after the merge is that the increased HBIOS size is not problematic for any of the standard builds.  Probably won't be.

Thanks, Wayne

kiwisincebirth:
Updated Testing
[Results of Testing HBIOS Function.docx](https://github.com/user-attachments/files/17271738/Results.of.Testing.HBIOS.Function.docx)

PR #431, opened 2024-10-07
RomLdr.asm now uses HBIOS SYSGET DIOMED to discover Slice Sector offset

kiwisincebirth:
RomLdr.asm now uses HBIOS SYSGET DIOMED to discover Slice Sector offset
See Issue #427 which this is part of the fix.


wwarthen:
This looks great Mark.  Thank you, I will merge shortly.

Thanks, Wayne

PR #432, opened 2024-10-08
Improvements to HBIOS

kiwisincebirth:
Improvements to HBIOS 
* SYSGET DIOMED : If Slice > 0 for NON High capacity drives now returns an error
* SYSGET DIOMED now returns device attributes in Reg B 
* Documentation improvements


kiwisincebirth:
Just Some Comments (on the above 3 improvements (in order)
* As discussed
* I added this in the to simplify the CBIOS code, I need to determine if the media is a floppy. I considered using the existing Media ID returned value, but comparison becomes problematic/buggy if anything changes. The Device attributes are for more deterministic from a comparison perspective, and I don't see any major downside?
* I hope the improvements (new section) make sense, happy for you to reword / add to it.

kiwisincebirth:
I tested the Rom loader boot function using ROM disk boot (don't have floppy disk) 
* "1.1" give a slice error
* "1.0" gives no system image 
So this looks good to me. 

I also updated my test harness and observer the new "Device Attributes: being returned. And I get the correct value being returned, but I am Not sure the value itself is correct ie. This is the value being returned correctly to me (from sd.asm)

SD_DEVICE:
	LD	C,%00110010		; C := ATTRIBUTES, REMOVABLE, SD CARD

but unsure why Bit 6 "Removable" = 0, shouldn't Bit 6 = 1 ? 

I search other drivers and they have similar issues.

wwarthen:
Thanks @kiwisincebirth.

The changes to SYSGET DIOMED looks good.  I can see where returning the device attributes is useful and I am fine with that.

Thank you for the documentation updates -- all good.

Regarding the removable bit for the SD driver.  You have uncovered a conundrum I ran into a while back.  The removable bit is used to determine drive letters will be automatically allocated for a device at boot even if there is no media in it (think of a floppy drive).  Even though devices like SD, IDE, etc. are technically removable, most users do not want drive letters auto allocated to them if there is no media inserted at boot.  In a perfect world, there would be a separate indicator for "auto assign drive letter".  There are no bits left in the device attributes byte and I have not come up with a good solution for this.

Your changes all look good.  Merging now.

Thanks, Wayne

PR #433, opened 2024-10-09
cbios changes to use HBIOS for Drive Slice Selection

kiwisincebirth:
cbios changes to use HBIOS for Drive Slice Selection

Relates to Issue #427


kiwisincebirth:
See the Testing Results in attached document.

ALSO NOTE : This change reduced the size CBIOS code by almost 0x100 bytes, (RomWBW) see the attached document for comparison. However I have NOT made any changes to recover this unused space. The space has been allocated to the working buffers in CBIOS. I leave this up to @wwarthen - to make appropriate adjustments. Noting that for UNA there has not been any realistic change in the code size, thus UNA and HBIOS variant(s) have diverged further in overall code size.

wwarthen:
Hi @kiwisincebirth,

In my cursory testing, I am seeing an issue with access to the ROM drive. I am seeing invalid directory entries instead of a normal directory.  Access to physical drives seems fine.  I will need a bit of time to look into this.  If you get a chance, let me know if you are having any issue with ROM drive usage under your changes.

In theory, there is no need to test CPM.SYS because the CBIOS component of it is binary identical to ZSYS.SYS.

Regarding the space that this change has freed up.  All of the people that I have heard from are looking for more disk buffer space rather than TPA.  As it is, this change frees up disk buffer space, so I will be leaving that as is for now.  Your change allows for mapping all drive letters A-P (in most cases).  Previously, this would result in disk buffer overflow.

Thanks! Wayne

wwarthen:
OK, I found the issue.  If DIOMED returns early (non-HD), the Media ID was not set properly.  I will merge this PR and follow it up immediatley with the fix.

Thanks! Wayne

kiwisincebirth:
> In my cursory testing, I am seeing an issue with access to the ROM drive. I am seeing invalid directory entries instead of a normal directory. Access to physical drives seems fine. I will need a bit of time to look into this. If you get a chance, let me know if you are having any issue with ROM drive usage under your changes.

FYI : Testing now without your commit, I dont have any issues accessing the ROM Drive, ill explore further

> OK, I found the issue. If DIOMED returns early (non-HD), the Media ID was not set properly. I will merge this PR and follow it up immediatley with the fix.

I looked at your commit, the path you have found is when MediaID=0, not non-HD media. when no media detected the Media ID returned was not set correctly. While still an issue, not sure its the one you described.

PR #434, opened 2024-10-10
Minor tweekas after last commit

kiwisincebirth:
* Moved DIOMED_MID variable - so it would be overwritten at start of main Calculation
* Removed code in SYS_GETDIOMED1A - since SYS_GETDIOMED6A did exactly this anyway
* Removed code that set MID_HD - since it would now already be set.

Relates to and improves on commit #433 - See Issue #427


wwarthen:
Excellent.  Thanks @kiwisincebirth

PR #437, opened 2024-10-14
Disable LCD as this board does not support it.

wdl1908:
Disable LCD as this board does not support it.


wwarthen:
Thank you @wdl1908!

-Wayne

PR #438, opened 2024-10-15
Fix: If the Slice=0 and No CPM partition

kiwisincebirth:
Followup Fix for Fusix boot issue. See follow commit. 

Issue is that Booting Slice 0 on Non-CPM Partition Disk, should ignore all Slice Calculations and Capacity Checks. Instead should just boot the Disk from LBA=0

https://github.com/wwarthen/RomWBW/commit/f3e0de4a97fca15f69878043031d8c06ed7db457

Primary Fix is on Line 5064, the rest is just code formatting.


kiwisincebirth:
@wwarthen - Sorry for introducing another issue - I tested this Does boot Fuzix.

wwarthen:
Thanks Mark.  I am out of time today -- will merge this tomorrow.

-Wayne

PR #439, opened 2024-10-15
CPM3 Loader Using New HBIOS SLICE Function

kiwisincebirth:
CPM3 Loader Using New HBIOS SLICE Function

See Issue #427


kiwisincebirth:
This has been tested both installed on media used SYSCOPY, and via the COM file (below), which is the main way to test the failure cases, as running it from the system track, the validation has already occurred, via the HBIOS boot loader.

```
14:36 A0:SYSTEM>fdisk80

HBIOS unit number [0..5]: 3
Capacity of disk 3:  (498M)  1019088      Geom 0f8c1010
Nr  ---Type- A --      Start         End   LBA start  LBA count  Size
 1     FAT16   06    260:0:1  1023:12:16       66560     952528  465M
   Reserved 4 x 8Mb CP/M slices

14:33 A3>zpmldr

Boot CP/M 3 from Disk Unit: 3   Slice: 9
\r\n\r\n*** Slice not supported
```

PR #442, opened 2024-10-17
Publish the CPM2.2 CPM.SYS Binary Files in a new Binary/CPM22 folder

kiwisincebirth:
Publish the CPM2.2 CPM.SYS Binary Files in a new Binary/CPM22 folder

Motivation.
a) Consistent with other OS (e.g. CPM3) publishing binary files to /Binary
b) During an upgrade will need these to update CPM22 boot drives. Currently only in Source folders

Testing: I have tested with Make, but don't have Win environment to test CMD files

### DIFF
--- a/Binary/CPM22/Clean.cmd
+++ b/Binary/CPM22/Clean.cmd
@@ -1,0 +1,4 @@
+@echo off
+setlocal
+
+if exist *.sys del *.sys
--- a/Binary/CPM22/Makefile
+++ b/Binary/CPM22/Makefile
@@ -1,0 +1,7 @@
+TOOLS = ../../Tools
+MOREDIFF := $(shell $(TOOLS)/unix/casefn.sh *.spr)
+
+include $(TOOLS)/Makefile.inc
+
+clean::
+	@rm -f *.sys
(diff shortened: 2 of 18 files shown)


wwarthen:
OK, sorry, I see that you did include the UNA variant for CP/M 2.2.  So I guess all that would need to be addressed is ZSDOS.

-Wayne

kiwisincebirth:
@wwarthen - I have added a commit to also publish the ZSDOS - SYS files. 

When I initially committed I didn't consider ZSDOS, I was just plugging a gap on CPM22, although it is good to fix both. I am fairly sure these SYS files will also be needed when upgrading, as they are linked to the BIOS Version number (cbios.asm).

kiwisincebirth:
Have now included QPM as well.

wwarthen:
Thank you @kiwisincebirth.  This looks straightforward, but I am going to run through some build tests to confirm.  Won't take long and I will merge after that checks out.

By the way, the reason that the CP/M 2.2 (and ZDSOS) system files were not previously included in the Binary folder is because they are included on the ROM disk.  However, I still agree that they should be in the Binary folder after a build as well.

Thanks, Wayne

wwarthen:
My testing looks fine.  Merging now.

Thanks, Wayne

PR #443, opened 2024-10-18
msx-keyboard: corrected implementation of msx-keyboard's tick hook

dinoboards:
Hi Wayne,

Found an issue with the MSX keyboard and the calling (or rather the not calling) of its `MKY_INT` handler.

If I configured a build to use `EZ80TMR_INT` for timer ticks, the `MKY_INT` handler would not be wired - and so keyboard would have 'issues'.

I noticed that the MSX keyboard's implementation for hooking into `HB_TICKS` handler was inconsistent.  There were 2 places the `MKY_INT` handler was called from.  And 1 was impossible to 'activate' - as the macro tested the conditional state, before it was `.SET`.

I have updated the `mky.asm` keyboard driver to inject itself into the `VEC_TICK` chain - i am assuming this is the intention here.

Dean


wwarthen:
Thank you @dinoboards.  I have another pull request queued up that I need to check out first, then I will merge this.  I don't have my MSX keyboard currently hooked up, so I am going to go ahead and merge this as long as there are no build errors.  I will do some testing in the next day or two.

Thanks, Wayne

Issue #440, opened 2024-10-15
Disk Catalog PDF Doc Improvement


kiwisincebirth:
I would like to raise an issue (enhancement) to improve the documentation for the project specifically in the Disk Catalog PDF document. This is primarily since I have put some time into this already, and think it can be greatly improved.

What is the Issue
* Firstly it is a bit incomplete, i.e. not all disks are described, 
* A lot of files and or descriptions are missing descriptions
* Contains duplicate file information, which makes it harder to maintain this document

Sources for Information
* The "d_*" directories that actually contain a list of the files
* The individual readme files in the "d_*"  directories actually contain a lot more information about the files than the PDF document, this is a big source of information
* The "hd_*.txt" files that contain definitions of what is copied onto the disk

Improvements
* Add sections for disks not already described, primary the standalone disks, e.g. d_games
* Create new section(s) which describes common files (e.g. Sound Files, Test Files, SIMH, etc) that are copied to multiple disks
* Remove the duplicate files from the main OS disk listings, and replace with pointers to the Common File sections.
* Copy disk and file descriptions from the individual readme files

Any comments, on how this could be improved. I will attach a sample showing where it is up to.

Mark.

[RomWBW Disk Catalog.pdf](https://github.com/user-attachments/files/17373026/RomWBW.Disk.Catalog.pdf)

wwarthen:
I strongly concur with the sentiment that the RomWBW disk catalog needs substantial improvement.  There have been multiple prior attempts to clean this up.  Sadly, none of them have "stuck".

From my point of view, the biggest issue is finding a way to keep the catalog up to date.  Improving the content of the catalog is probably just work.  Finding a process that makes maintenance of the catalog easier and more foolproof.

I also wish to solicit thoughts and ideas.  Improving the disk catalog would greatly enhance RomWBW.

Thanks, Wayne

kiwisincebirth:
@wwarthen You also made a comment in the document itself:

```
**NOTE:** This document has become somewhat out-of-date.  I apologize
for this.  It is very difficult to maintain due to the volume of
disk content updates and the way that files are dynamically merged
into disk images during the build process.
```

One of the primary changes I have made is to break up the content of the disk catalog into various sections, which are referenced , thus there should only be ONE place where a file is defined. e.g. "FDISK80.COM" appears once, previously it appeared 5 times. Thus : 

This should hopefully improve the maintainability of the document

[...]

PR #444, opened 2024-10-19
Fixes to Imgae file Defs, and move test files, and readme files

kiwisincebirth:
Made correction to move some general purpose application files to user Area 15, consistency with the OS convention.

Moved Image/Test into Image/Common/Test - didn't see the point for it to be high level, as it is not a disk image itself, it is a common set of files shared across most disk images

I corrected renamed ROMWBW.txt (U0) in cpm3 and zpm3 -> to Readme.txt file and moved to the parent directory to be consistent with other OS and then include it in the disk image with the consistent readme.txt filename


kiwisincebirth:
Relates to issue #440 - These changes were made after general review of the documentation.

PR #447, opened 2024-10-23
Merged /Common/Utils into Common/All, and reorganised.

kiwisincebirth:
Changes
* Moved all /Common/Utils into /Common/Apps
* Reorganised /Common/Apps, creating U10 U14 U15 directories
* Corrections to all disk definitions
* Omitted U10 on space constrained FD images

Note: Some FD image definitions have had # commented  lines added back to highlight potentially missing files that could potentially be added back. But I think this need to be more consistent across all floppy definitions.

See Issue #440


wwarthen:
Thank you @kiwisincebirth.  Looks great.

wwarthen:
Thank you @kiwisincebirth!

Issue #451, opened 2024-10-27
Front Panel Switch Enhancement


kiwisincebirth:
Motivation: When I boot i have to type the name of the disk and slice.
and 90% of the time it is the same. While I can customise the ROM
and set a default, this is not very user friendly.

Alternatives: Front panel switches, while posible, does reqire
additonal hardware, and have found drive selection not very flexible.
i.e. I have my OS's on Slices 10-15, and data on slices 0-9, this
means I get all data drives mapped consistently starting at D:
an a single os drive mapped to A: Clearing booting slices 
above 7 is not possible using switches, so wont work.

## Idea:

In Systems with RTC that has NV RAM store the FP Switches in NVRAM
replacing need for physical switches

Extend the Switches themselves from 8 -> 16 bits, to store additional
data to improve the boot experiance. 16 bits works well with current 
API's - that return switches in L, and leave H = 0, which in souce code
a comment says for future. Potentially physical switches could extend 
to 16 bits also, if really needed.

Additional switches 8->16 could be used for

* Specifying Timeout for Auto Boot, (curenlty there is no timeout for switchs)
* Improved Disk/Slice selection
* amnything else ?

A CP/M app, (or ROM app) to change switch data in RTC NVRAM would be required

During boot (romldr) RTC NVRam would be read, and if detected would use
the switches to :

2 Modes of operation during boot

* Auto Boot - "with timeout" - will boot using switches if timeout occurs
  Menu is displayed, if timeout > 0
* Manual Boot - Defaut option (pressing <Return>) comes from Switch data
  currently default comes from ROM (configured during build)

If RTC NVRAM is present, and switches esist in NVRAM, then
The above Replaces the the use of the following in ROM build config.

* BOOT_DEFAULT - string - Typically "H"
* BOOT_TIMEOUT - seconds before default boot occures.

Otherwise the above values would be used as they are today.

### H Register (New) contains
Bit 7 - = "1" - Virtual/Extended front panel, identifies 16 bit switches
Bit 6 - Reservered - Future
Bit 5 - Reservered - Future
Bit 3-4 - BOOT_TIMEOUT in seconds (0-3) 0=immediate, applies to auto boot
Bit 0-2 - Disk Unit to Boot (0-7), 8 drives should be enough?

### L Register (Changed) is:
Bit   7 - (unchanged) CRT/SER CONSOLE
Bit   6 - (unchanged) SEC/PRI CONSOLE
[...]

PR #455, opened 2024-11-15
Map/switches2

kiwisincebirth:
Support for ROM Loader using FP Switches - Relates to #451


wwarthen:
Hi @kiwisincebirth,

Just letting you know that I am now back in town and starting to review this PR.  Not sure how long this will take me, so wanted to let you know I am working on it.

Thanks, Wayne

wwarthen:
Hi @kiwisincebirth,

It has continued to be a stuggle to find time to work on this, but I am getting there.  The builds (Windows and Unix/MacOS) seem to be fine and I have been able to test a variety of scenarios with no obvious problems.

Some notes so far:

- sysconf.com is being built in Windows, but not Linux/MacOS.  I will remove it from Windows for now in order to get binary identical builds.
- The documentation is very good, but I do have a few things to add.  I will take care of this and you can review aftrerwards.
- I am thinking of adding a message in RomLdr like "AutoBoot in 5 seconds..." before the prompt when AutoBoot is active.  This would apply to either NVRAM or custom ROM settings.
- All of the standard ROM configurations build OK.  However, there are some commonly used configurations that are now failing due to HBIOS space.  This is not really due to your enhancements -- there are multiple things contributing to this.  I'm not sure I can release the current code tree until I work through some of the space issues.

If everything continues to look OK, I may merge your PR later today.  When I do the merge, I will follow up with my minor changes.

Thanks! Wayne

kiwisincebirth:
> The documentation is very good, but I do have a few things to add. I will take care of this and you can review aftrerwards.

I am sure your changes will be OK, ill review. I tried to capture everything that was discussed, and to not duplicate information between the documents.

> I am thinking of adding a message in RomLdr like "AutoBoot in 5 seconds..." before the prompt when AutoBoot is active. This would apply to either NVRAM or custom ROM settings.

FYI I already a message for detection of NVRAM see. `str_nvswitches`, but this is different.

> there are some commonly used configurations that are now failing due to HBIOS space

FYI I only primarily build for SCZ180/sc_126 : In terms of space that is probably a whole other ticket to look for optimisations, I can see some minor optimisations that may save a few bytes here and there. However one bigger one might be to look a consolidation of the DSKY autoboot, and the autoboot from command line, there is definately duplication in the logic. Again a bigger topic

> Based on your documentation, it looks like your intention was to build sysconf.com and put it in /Binary/Apps.

Originally sysconf was built independently of HBIOS as both ROM and COM apps, I left the com file generation (didn't see point removing it). There is the case whereby if you set autoboot with timeout = 0, it bypasses user input. Having sysconf available as a com app, allows this to be changed (without removing RTC chip or battery)

> That is fine. So, I will add it to the Linux/MacOS Makefile and leave it as-is in the Windows build.

Thanks for resolving this.

wwarthen:
Hi @kiwisincebirth,

I am going to go ahead and merge this now.  I will follow-up with a minor commit of my own.

In my testing/review, I am seeing one area that needs to be revisited.  When NVRAM is valid, but ABOOT_AUTO is OFF, the ROM-based auto boot processing is bypassed.  I think that if ABOOT_AUTO is OFF, then processing should revert to the the ROM-based autoboot processing.  Right?

I don't consider this a showstopper for the merge at this point, but can you look at this?  Or let me know if I have misunderstood something.

Thanks! Wayne

kiwisincebirth:
@wwarthen 

> In my testing/review, I am seeing one area that needs to be revisited. When NVRAM is valid, but ABOOT_AUTO is OFF, the ROM-based auto boot processing is bypassed. I think that if ABOOT_AUTO is OFF, then processing should revert to the the ROM-based autoboot processing. Right?

Its an easy fix. Change this statement `JR	Z,prompt` to say `JR	Z,nonvrswitch`. (line 295)

However it may not be what you expect (exactly). For background and from my understanding there are 2 independent variables, functionally speaking.
* default boot controls the string that is processed when a user presses <enter> at the boot prompt, set to "H" by default. in rom config and in nvram
* auto boot control if and when <enter> is pressed automatically at startup.

it is one of the reasons i broke it into 2 NVR switches, to align to the 2 rom config options, breaking with single 8 bit value for FP switches

The only slight issue if change above is made is the config could come from two sources and this might be confusing. e.g. a user sets (in config.asm) a boot device and auto timeout, but when booting the default boot would come from NVRAM, once configured NVRAM get the "H" boot default, an this always takes precedence, there isnt currently a way to say ignore nvram boot default. 

Does this make sense ?

IMO If NVR is configured, it should take precedence, effectively leaving rom config only for non-nvr enabled systems keeps it simple, and easy to explain. - Just my opinion

I leave it with you to make the call.

wwarthen:
Hi @kiwisincebirth,

Yes, you are correct that the autoboot command line is used in 2 ways: 1) if autoboot is enabled and the timeout expires, or 2) if the user presses <enter> with an empty command line.  This is confusing and was ill conceived from the start.  Not sure what I was thinking.  I am going to completely remove the concept of the autoboot command line being used as a default if the user enters an empty command line.

I also see your point about NVR taking precedence over the ROM config if it is valid.  If I make the change to remove the concept of the default command line, then everything should be OK.

I will work on this.

Thanks, Wayne

wwarthen:
Hi @kiwisincebirth,

OK, I am about to check in some changes.

I have removed the whole "default command line" concept.  Very happy with that change, but I still need to change the config files to refer to this field as "AUTO_BOOT" instead of "BOOT_DEFAULT".  That will be in a subsequent check-in.  I need to do something tricky to avoid breaking configs that still use the obsolete field name.

I have changed my mind yet again about how to handle ROM config if NVR is valid.  Sorry.  With this check-in, if NVR ABOOT_AUTO is OFF, then processing **will** proceed with ROM config.  I am still on the fence about this, but doing it this way is making my testing a lot easier for now.  Both options are in the source code (with one commented out) and comments describing the situation.

Thanks, Wayne

PR #460, opened 2024-12-04
Master

b1ackmai1er:


wwarthen:
Hi @b1ackmai1er,

Thanks for submitting this.  I am finishing up some work on another PR that I want to finalize before introducing this.  Please give me a little time to get to this PR.

Thanks, Wayne

Source/HBIOS/kbd.asm
```diff
@@ -59,14 +66,29 @@ KBD_IDLE	.DB	0	; IDLE COUNT
...
+#INCLUDE "ps2iface.inc"
+;__________________________________________________________________________________________________
 ; KEYBOARD INITIALIZATION
 ;__________________________________________________________________________________________________
 ;
 KBD_INIT:
+#IF ((INTMODE == 2) & INTPS2KBD)
+	CALL	KBDQINIT		; INITIALIZE QUEUE
+	LD	HL,KBD_INT		; INSTALL VECTOR
+	LD	(IVT(INT_PS2KB)),HL	; IVT INDEX
```
wwarthen, line 81: I don't think the interrupt vector should be installed until after you have confirmed the existence of the keyboard controller.

Source/HBIOS/ps2iface.inc
```diff
@@ -0,0 +1,247 @@
+;__________________________________________________________________________________________________
+; HARDWARE INTERFACE FOR POLLED AND INTERRUPT DRIVEN PS2 KEYBOARD.
+;
+; IN POLLED MODE THE KEYBOARD DEVICE IS IDENTIFIED BY THE TABLE POINTED TO BY IY
+; IN INTERRUPT MODE THE KEYBOARD DEVICE IS UNKNOWN SO PREDEFINED PORT VALUES ARE USED.
```
wwarthen, line 5: It would be preferable to use IY in this file and set IY in the interrupt routine itself.  Take a look at sio.asm starting at line 245.  If only one KBD device can exist with interrupts enabled, then IY should still be set at the start of the interrupt handler and IY should be used in this file to get the port numbers.

Source/HBIOS/ps2iface.inc
```diff
@@ -0,0 +1,247 @@
...
+;
+;__________________________________________________________________________________________________
+; KEYBOARD INTERRUPT VECTOR
+;__________________________________________________________________________________________________
+;
+KBD_INT:
+;	CALL	KBDQDBG
+	LD	A,(KBDPORTS)		; READ CONTROLLER STATUS
+	LD	C,A
+	IN	A,(C)
```
wwarthen, line 73: It looks like the EZ80_IO macro is missing from the new I/O calls.

wwarthen:
Hi @b1ackmai1er,

Did a cursory review and had some questions.  They are posted in the PR conversation.  Can you take a look?

Thanks, Wayne

Source/HBIOS/ps2iface.inc
b1ackmai1er: Thanks Wayne.  I didn't know there was a way to resolve this. Will take a look.

Source/HBIOS/kbd.asm
b1ackmai1er: Thanks Wayne I think I understand what I need to fix this.

Source/HBIOS/ps2iface.inc
b1ackmai1er: Fixed. Thanks.

wwarthen:
Hi @b1ackmai1er,

Thank you for addressing my questions.

I am trying to get the code working on my system, but not having much luck yet.  I strapped the PS2 interrupt to int #2.  I think this is working because I get bad interrupts on int #2 after doing this and still using old code.  When I then switch to your code and enable INTPS2KBD, I see the interrupt line go active, but does not ever seem to be serviced.

I am positive I have IM2 interrupts working because my timer is working.

Any ideas?  I assume the current code is working for you.

Thanks, Wayne

wwarthen:
@b1ackmai1er, never mind.  I just figured it out.  I had the jumper on the USB interrupt instead of the keyboard/mouse interrupt.  Working now.

Thanks, Wayne

wwarthen:
Hi @b1ackmai1er,

So, I struggled for a while with this.  After I got it working, I found that every time I restarted the system, I would get a stream of BAD INTs.  I finally found that the 8242 controller was defaulting to interrupts enabled and started interrupting the system before the interrupt vector was put in place.

I was able to get this under control by adding an ugly hack to send a command to the 8242 to disable it's interrupts prior to interrupts being enabled globally.  Later on, the 8242 is reprogrammed to generate it's interrupts.

Anyway, I think I have things stable at this point and intend to check in your PR tomorrow (followed by my tweaks).

I have to admit I am not entirely sure the added complexity and code space of this is buying us a lot.  The PS2 keyboards have a 16 byte buffer built-in.  Do you have a use-case for this?  How strongly do you feel this is valuable?  I'm OK to add it, just asking.

Thanks, Wayne

b1ackmai1er:
Yes I have been struggling with the keyboard interrupts enabled issue as well. This was not an issue for me initially but was struggling to understand and resolve it. I think it is often masked by previous keyboard controller configuration and warm boot vs doing a cold/power on boot. I don't have a usage case. It was just a opportunity to implement the full hardware capability.

b1ackmai1er:
@wwarthen 

So my understanding at the moment is that interrupts are not active for the PS2 controller at cold boot. I believe doing a controller self test to identify if the controller exists enables interrupts prior to getting to the initialization section to enable interrupts. That is why we only see the interrupt errors in the middle of the driver initialization message. I think we need to disable interrupts while doing that test.  Will explore this.

wwarthen:
Thanks @b1ackmai1er, your fix for the spurious interrupts looks much better than the hack I used.  Let me test your latest and confirm.

Also, I am working on rearranging a few things to try and reduce the complexity and size overhead of this.  I do want to include this enhancement, but also want to do everything possible to keep HBIOS size and complexity down.  Give me a little time to work through this, then I will have you review my updates.

A big part of the problem is my original design for the keyboard drivers and the way that I have imbedded them in the video drivers.  It was an ill-conceived approach that I eventually need to resolve.  Keyboard and video drivers should be separated and "hooked together" in a different way by the terminal code.  Someday...

Thanks, Wayne

wwarthen:
Hi @b1ackmai1er,

I am going to go ahead and merge this PR.  I will be following up with my own commit that tweaks a few things.  I then intend to review the code a bit more because I think there is code included in the "non-interrupt" build scenario that is not needed.

Regarding the spurious interrupts.  Your fix does basically seem to work.  However, I can make it break if I pound on the keyboard during the boot process.  If there is already stuff in the output queue of the 8242 at boot, then it asserts an interrupt long before the keyboard initialization takes place.  As long as you leave the keyboard alone during boot, it is fine and I am OK with that for now.  Really not sure how many people will even turn this on.  I did find a way to avoid the spurious interrupts completely even if the keyboard is pounded on during boot.  You will see my code for that when I commit, but it is commented out.  It is a real hack and I don't want to include it.  I intend to just remove my hack completely at some point.

Thanks, Wayne

wwarthen:
OK, I am done mucking about with this.

@b1ackmai1er, let me know if you see any issues at this point.  Seems to be working well for me.

Thanks, Wayne

PR #467, opened 2024-12-09
PS2 keyboard spurious interrupts

b1ackmai1er:
Resolved keyboard interrupt errors when a key is pressed during reboot. Turn off keyboard controller interrupt enable before cold boot process starts. I'm wondering if  duodyne interrupt generator is immune to DI :)?


wwarthen:
I am currently dead in the water.  Power outage.  May be another 3 days.

wwarthen:
Hi @b1ackmai1er,

The refactor generally looks good.

I do have some reservations about handling the spurious keyboard interrupts via a de-initialization routine:

- It introduces a exception to the generic driver module (only applies to KBD driver).  I am trying to avoid such exceptions.

- While I realize it works, it is considered best practice to address device reset at boot rather than at shutdown.  This is because you can't absoloutely guarantee how a system is being reset.  I know this is a bit esoteric, sorry.

Anyway, I propose to merge your change, but then convert the de-initialization mechansim to an initialization mechanism.  This would mean adding the video drivers to the pre-init call table.  The TMS driver already has this and all of the video drivers probably should have been this way from the beginning.  It is consistent with the existing driver model.

Are you OK if I do that?

Thanks, Wayne

wwarthen:
Hi @b1ackmai1er,

I have merged your changes.  They worked fine.

I have subsequently made the change to disable interrupts in PREINIT.  It seems to be working well for me.  I have not been able to generate any spurious interrupts.  Can you please test?

Thanks, Wayne

PR #475, opened 2024-12-16
Roll out kbd preinit into cga, gdc and vga driver

b1ackmai1er:
Don't beleive this is required for vrc


wwarthen:
Looks great @b1ackmai1er.  Thank you!  I will merge now.

I did notice that the CVDU driver was not enhanced with a PREINIT to call KBD_PREINIT.  That is trivial to add and I will take care of it.

Thanks, Wayne

PR #479, opened 2024-12-18
Changed the Attribute for MD RF devices

kiwisincebirth:
Changed the Attribute for MD RF devices 
reordering/flipping (RF<->Flash) so bit masks are easier

See Discussion #477


wwarthen:
Thanks @kiwisincebirth!

-Wayne

PR #480, opened 2024-12-20
Added Disk Labels (Loader.ASM)

kiwisincebirth:
This just updates the disk labels across the 5x loader.asm files.

Note: i made some code improvements to the UNA loader code to correctly use .EQU 's defined, (a) removing duplication of constants, and (b) better exposing sector count (22) in its own .EQU

And: I also corrected CPM3/ZPM3. While the code was coped from CPM22 it was never updated for cpmldr.bin, yes I know the loader is not used. However I have made same changes to it for "consistency"


wwarthen:
Thank you @kiwisincebirth!

PR #482, opened 2024-12-21
CBIOS Size reduce by reducing resident utils

kiwisincebirth:
three things (a) removed unused utils (b) moves utils to startup non-resident code that was not needed generally (c) small LD BC, optimisation when calling HBIOS.


wwarthen:
Excellent work @kiwisincebirth!

Thanks, Wayne

PR #484, opened 2024-12-22
Consider heap space remaining before each ASSIGN /B= assignment

kiwisincebirth:
NOTE : This was actually tested with a CBIOS with a .FILL of $800 bytes (in resident code) to reduce heap size, and assert that allocations stoped before running out of heap. 

I left a couple of debug statements in code to allow the remaining heap estimate to be displayed, and during testing the estimates were very slightly different (few bytes), compare to the reported bytes remaining, but not enough to make any difference.


wwarthen:
Hi @kiwisincebirth,

This looks pretty good.  It will definitely be a little off for RAM/ROM/Floppy drives in some cases because they have different sizes depending on system configuration.  Regardless, this is probably fine for the purpose.

Thanks, Wayne

PR #486, opened 2025-01-04
Option to move Fonts out of HBIOS area to ROM Bank 3 

b1ackmai1er:
The FONTS_INLINE equate controls whether the fonts are included in the HBIOS area as per existing behaviour or whether they are placed in the currently unused ROM bank 3.

In the case where we use bank 3 for font storage, the HBIOS is built with the font data located at 0x8000 onwards. The build process splits this off for relocation to the correct image position.  Unlzsa, the TMS and VGA driver have been updated to access the font data from either location, whether compressed or not.

Moving the font data to bank 3 results in significant savings in HBIOS space.

If you are happy with this approach, I will continue developing and roll changes into other display drivers and do some hardware testing.

Further on, as mentioned, I would like to look see if i can get the decompressor to decompress one byte at a time and move the CCP cache to bank 3 as well.

Regards Phil


Source/HBIOS/Build.cmd
```diff
@@ -123,6 +123,12 @@ if %Platform%==S100 (
 
 copy /b romldr.bin + dbgmon.bin + ..\zsdos\zsys_wbw.bin osimg_small.bin || exit /b
 
+::
+:: Split off anything above 32K mark of the hbios. This will be relocated to rom bank 3
+::
+
+"%TOOLS%\srecord\srec_cat.exe" hbios_rom.bin -Binary -crop 0x8000 0x10000 -offset -0x8000 -o osimg2.bin -Binary || exit /b
```
kiwisincebirth, line 130: An opinion:

This command overwrites and replaces the logic (116-122) that creates osimg2.bin. This will probably break S100 functionality, It would be easier to understand if you included this line in the above if statement

I don't see any conditional IF FONTS_INLINE around this code, should there be?

The concept of building an oversize hbios.bin file which you then split-up using this command, is counter to the approach used in much of the build. i.e. Generally should build smaller BIN files specifically with the content you need and append these files together using binary copy. better to stick to this approach for clarity and consistency

I didn't see any changes to makefile to support this change

wwarthen: Hi @kiwisincebirth,

I have been having an offline conversation with @b1ackmai1er about this.  This conversation probably should have been as part of this PR or a discussion in the repository.  Sorry about that.

I had already suggested to @b1ackmai1er that the imbedding of the fonts in the alternate ROM bank be done differently.  Specifically, I am advocating that all fonts be imbedded and indexed by a simple table of font id's and pointers.  This would allow the fonts to just another chunk within a ROM bank (like all the other .bin chunks).

Thanks, Wayne

wwarthen:
Hi @b1ackmai1er,

I apologize for taking so long to get back to you on this.  I have been challenged with multiple things going on here personally and have had almost no time for this project.

You have created some excellent code that I do want to include in RomWBW.  However, I am really feeling like this is more change than I want to deal with right now.  I am trying to lock down the source for v3.5 and get a new stable version released.  I would prefer to include this work in v3.6.  This means it will be on hold for at least a month and I apologize for that.  However, I have committed to a new stable release by mid-February and I really need the next few weeks for testing with a stable codebase.

Finally, as we have discussed offline, I do want to modify the way that fonts are imbedded in the ROM bank.  I want to create a new assembly module that begins with a table that has an index into the fonts followed by the font data itself.  I know how I want to do this, but I don't want to take the time to work on it right now.

Thanks, Wayne

wwarthen:
Closing this PR because I think I have incorporated all of these features in recent updates.

Thanks, Wayne

Issue #489, opened 2025-01-15
Build completion with oversized Rom Bank 1


kiwisincebirth:
I discovered an issue (last year) whereby increasing the size of ROMLDR ( while testing of new menu option for "Slice Inventory" ) , I was actually able to increase its size build and deploy it to a ROM without any issue, when in fact the Rom Bank 1 , containing oversized RomLdr would wpuld truncate the bank, and probably (didn't specifically test this) fail to load the CPM from the Rom, since CPM (last in bank) would be truncated

I have developed a simple fix the takes a similar approach to Rom Bank 2, and will generate a "Build Error" if Rom Bank 1 is oversized. I have also moved this logic out of usrrom.asm (for Rom Bank 2)  into layout.inc, which contains the oversize detection logic for both of these banks.

Will create a PR if this makes sense?

wwarthen:
Yes, this sounds pretty straightforward and worth including in v3.5.  I will want to review the code first, but go ahead and generate the PR.

Thanks, Wayne

kiwisincebirth:
closing as resolved by pr

PR #491, opened 2025-01-16
Rom Bank Layout improvements

kiwisincebirth:
Issues
* Rom Bank 1 has no size constraint, can increase romldr (oversize) without build issue, bank truncation 
* Rom bank 2 usrrom (last app) has size constraint check, shouldn't be function of app 
* Rom bank 2 usr rom pads to end of 32k in its Size, unnecessary as handled by "srec_cat -crop"

So
* move out the bank size protections (from usrrom) into layout.inc
* ensure bank 1 has same protection as bank2.
* make user rom a realistic size ($200), padding handled by CRC check
* also good to .ECHO overall sizes of banks, improve logging.
* updated documentation in layout.txt

Addressed Issue #489

### DIFF
--- a/Source/HBIOS/Layout.txt
+++ b/Source/HBIOS/Layout.txt
@@ -5,34 +5,49 @@ Final Output Files
 ROM Output File [512K] -> <config>.rom
   hbios_rom [32K]
   OSIMG  [32K]
-  OSIMG1 [32K] 
-  OSIMG  [32K] (used as filler)
+  OSIMG1 [32K]
+  OSIMG2 [32K]
   romdisk - [384K]
 
-COM Output File -> <config>.com
-  hbios_com [varies]
-  OSIMG [32K]
+UPD Output File [128K] -> <config>.upd
+  hbios_rom [32K]
+  OSIMG  [32K]
+  OSIMG1 [32K]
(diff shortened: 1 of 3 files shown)


wwarthen:
Thanks for this @kiwisincebirth.

I do have one question related to removing the module size check from usrrom.asm.  If someone exceeds the available space in usrrom.asm, the behavior of the build will depend on whether Linux/Mac or Windows is being used.  Under Linux/Mac, you will get a reasonably nice message that the "SLACK" is negative.  However, under Windows, TASM just crashes with no reasonable message to the user.

I would be inclined to restore the size check in usrrom.asm.  Thoughts?

Thanks, Wayne

kiwisincebirth:
First off I hadn't considered differences in how the builds would display an error message, I was just following the existing mechanism for raising the error

Second I think the size check in usrrom should be restored, on reflection this was a mistake. Reason: the checks are independent of one another: usrrom.asm checks that that module is sized correctly and it doesn't its stated size; layout.inc checks the overall size of the bank is < 32kb

kiwisincebirth:
@wwarthen - latest commit restores code as per suggestion

wwarthen:
Looks good.  Thank you @kiwisincebirth!

PR #503, opened 2025-02-18
HBIOS UART - manage /DTR along with /RTS

feilipu:
The RC2014 [Single UART](https://8b8bf43264c2f150841a.b-cdn.net/wp-content/uploads/2024/12/16C550-UART-Schematic.pdf) and [Dual UART](https://8b8bf43264c2f150841a.b-cdn.net/wp-content/uploads/2024/12/16C2550-Dual-UART-v1.1.pdf) Modules do not have the /RTS signals(s) connected.
The /DTR is used to signal flow control to the attached FTDI USB adapter(s).

This PR simply adds active management of the /DTR.

The existing UART initialisation process already does /DTR configuration, but not active management.


wwarthen:
Thank you @feilipu.

Unfortunately, this will only work when interrupts are active for the 16C550.  Under most circumstances, the 15C550 will be capable of "auto" RTS/CTS flow control which is far more efficient than interrupts.  I'm afraid that Spencer's board is, therefore, incapable of handling auto-RTS/CTS.  Additionally, the UART driver currently requires interrupt mode 2 to do interrupt processing.  The RC2014 generally runs under interrupt mode 1.  I may be able to change that, but it is suboptimal.

I am going to ping Spencer and see if future revisions of the board can be changed to use RTS/CTS instead.

Thanks, Wayne

feilipu:
Thanks.

That's the right approach.

wwarthen:
Hi @feilipu,

Auto flow control is enabled by default on all UART ports that have support for it.  That is active in v3.5 (and v3.4).

Interrupt driven flow control (to which you added /DTR) can be enabled in the UART driver for chips that do not support AFC by setting the config variable UARTINTS to TRUE.  It is set to FALSE by default because enabling UART interrupts adds overhead to all ports in the UART driver -- even for chips that support AFC.

I have been interacting with Spencer on this.  He is going to test using UARTINTS for the existing boards to confirm it works.  If so, he will either provide the instructions for turning on UARTINTS or possibly just distributing a custom ROM with it already enabled.

Does this make sense?

Thanks, Wayne

feilipu:
> Auto flow control is enabled by default on all UART ports that have support for it. That is active in v3.5 (and v3.4).

Exactly what I was hoping.
🙏

PR #507, opened 2025-02-26
Source Code (Inactive) for CopySL Utility

kiwisincebirth:
Hi @wwarthen - Here is the source code for CopySL utility. Please Note:

* This adds the source code, I haven't changed the build scripts to compile and copy it, ill leave it when you ant to do this.
* I converted to to use TASM, and have tested a new COM file and it appears to work.
* Conversion from Z80ASM to TASM took a while until I worked out replaced ".DS" with ".FILL" directives. Other directives were changed to include the "." before the directive. Z80ASM doesn't seem to require it.
* When I copied the code I took the opportunity to combine the main program into a single file, previously it was broken up into multiple includes (TE editor size limitation), thus the output is not binary identical. ie. Includes at top of main file <-vs-> being pasted into the logical place in the file itself. I did put comment blocks where this pasting occurred.
* Apart from above no other code changes were made.

Any other questions let me know?

Mark.

PS. This code is my first serious Z80 code since the 80's, and was written on TE editor on the actual device (formatting is not great), so cut me some slack 😄

### DIFF
--- a/Source/Apps/copysl/cio.asm
+++ b/Source/Apps/copysl/cio.asm
@@ -1,0 +1,328 @@
+
+BDOS   	    .EQU 5
+
+; bdos commands
+CONIN       .EQU 1
+CONOUT      .EQU 2
+DIRCONIO    .EQU 6
+
+; TODO for more routines see assign.asm
+
+; ===============
+; INPUT
+
+; Console Input
+getchr:
+    PUSH    BC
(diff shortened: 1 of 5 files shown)


wwarthen:
Thank you @kiwisincebirth, merging now.

I intend to add this to the build process shortly.

Thanks, Wayne

PR #526, opened 2025-04-09
Added Updated SLR Tools with newer or missing versions

kiwisincebirth:
Based on a Posting on Google Groups I added the missing tools or updated to slightly newer versions.

Note: I don't have any documentation for the PLUS versions so not exactly sure what the differences are, hence updates to the documentation are fairly sparse, hence just put some placeholders.

https://groups.google.com/g/rc2014-z80/c/jc5eP6mB8tg/m/BglWGoX5AgAJ


wwarthen:
Very help.  Thank you @kiwisincebirth.

-Wayne

PR #545, opened 2025-05-15
Some enhancements to the pcf driver.

wdl1908:
Added default values for PCFCLK and PCFTRNS 
Added comparisons to set PCF_CLK and PCF_TRNS


wwarthen:
Thank you for this @wdl1908.

I need to be super careful with driver changes right now because I am kind of "between" releases.  The current RomWBW branch is v3.5.1 which is intended to be only minor patches and documentation improvements.

I am currently away from home, so it is almost impossible for me to test these changes even though I see they are straightforward.  I will be home in a few days and will look at this.  If the changes result in a binary identical build of the standard configurations (which they should), they I will proceed with the merge.

Note that I see a couple things I might do to make things a little more consistent with RomWBW, but I can do that easily after I merge your work.

Thanks, Wayne

wwarthen:
Thank you @wdl1908!

-Wayne

Issue #548, opened 2025-05-21
Add support for MS-COBOL-80


mggates39:
I am using my SC-131 to demonstrate several "Hello World" programs on CP/M.  One of those is COBOL.  To that end, I have added and MS_COBOL-80 disk image source to my repository and would like to issue a pull request to include it in the next release.  It is based on the MS COBAL v.4.65 archive found at [RetroArchive.org](http://www.retroarchive.org/cpm/lang/lang.htm)

I have the following:
Source/Images/d_cobol directory with a readme.txt file
Source/Images/fd_cobol.txt and Source/Images/hd_cobol.txt files
Doc/Language/Microsoft_COBOL-80_Manuals_1878.pdf file
Updated Source/Images/Build.cmd to include the COBOL disk image
Updated Source/Images/Makefile to include the COBOL disk image.

wwarthen:
This sounds great @mggates39.  I am working on finalizing the v3.5.1 release today and will be transitioning to v3.6 development.  We can introduce this disk image with v3.6.

You may generate a pull request whenever it is convenient.  I will do the merge after the v3.5.1 is published.

Thanks, Wayne

wwarthen:
This work has now been merged, so I am going to close this issue.

Thanks, Wayne

PR #549, opened 2025-05-22
Add MS-COBOL Disk image

mggates39:
Resolves Issue #548 to add the support for MS-COBOL-80 V4.65 to the collection of compilers available


wwarthen:
Thank you @mggates39!

-Wayne

PR #562, opened 2025-05-31
Use GFM version of ReadMe.md for MkDocs.

codesmythe:
This commit fixes the issue, discussed in #553, of the documentation home page showing a version of `ReadMe.md` that has not been preprocessed. Now `ReadMe.gfm` is displayed, which is preprocessed.


wwarthen:
Thanks @codesmythe!

PR #565, opened 2025-06-02
Add VDA driver for Xosera, an FPGA-based video controller.

codesmythe:


codesmythe:
Hi @wwarthen, this is the VDA driver for the [Xosera](https://github.com/XarkLabs/Xosera) RCBus module we discussed a few months ago. Everything about Xosera and my "port" of it to the RCBus is or will be open-source. I'm still working on getting the info about the module I've put together published. 

We had discussed using a base I/O address of 0xA0, but I could not get that to work in my RCBus system, so I've set it to 0x20 for now. I still need to investigate why 0xA0 does not work.

Thanks for including this, and for all your work on RomWBW.

wwarthen:
Hi @codesmythe,

Just wanted to let you know I tweaked the driver file.  There were numeric literals using the 0x prefix.  This seems to be OK with uz80as, but TASM chokes on it.  I globally replaced them with $ prefixes.

Thanks, Wayne

PR #569, opened 2025-06-06
Limit Xosera config to 2 UARTS; set $A0 base addr.

codesmythe:
The consenus is that the best default address for Xosera is $A0, so make that the default.

However, in that case we need to limit the number of UARTs that are probed to two because the probe for a third UART writes unlucky values to Xosera (at $A3) and causes it to reconfigure itself and lock up the bus for a time.

Thanks to Wayne for his help in debugging this.


wwarthen:
Thanks @codesmythe.

-Wayne

PR #570, opened 2025-06-06
Driver for PC style RTC, MC146818/DS1285/DS12885

mabartibin:
As discussed on the RC2014 forum, here is driver code for a DS12885 RTC, or its predecessors, with parallel interface.


mabartibin:
Oops, I managed to botch the DSRTC code in hbios.asm somehow.  Good you spotted that!

Since none of the usual Z80 machines uses this type of RTC, I wonder whether there should be documentation pointing to the hardware that would work with this code?  And where?

Issue #575, opened 2025-06-27
Slice Inventory App - Called from RomLdr


kiwisincebirth:
Create a New ROM application to display a List of Slices - invoked from the Boot Menu

This is a continuation of the discussion
https://github.com/wwarthen/RomWBW/discussions/477

Intention:
* Create a new rom app, a subset of code copied from `slabel.asm`, adding it to the newly enabled Rom Bank 3
* Invoked (differently) by a Call and Return to Boot prompt, rather than a Jp and Warm Boot that is prevalent in other more standalone apps
** Also : Invocation is coded as a built in application rather than generic Application Load
* App will be Hidden from 'L'ist rom apps, but displayed in the main 'H'elp menu
* Would NOT work in the App Boot Mode
* As per `SLabel`, does not currently support UNA bios - Code was never written as no way to test
* Currently expect it to easily fit in $200 bytes of ROM

kiwisincebirth:
@wwarthen Question

I was intending to use "S" character for the invoking. 

```
Boot [H=Help]: h
  D           - Device Inventory
  S           - Slice Inventory
```

However this conflicts with `#if (PLATFORM == PLT_S100) ra_ent(str_smon,  'S' ` - so not sure how what key to use?  Options

(1) Since S100 is a small percentage of hardware; Rename the `S`  key for S100 to something else e.g. `1`. I cannot think of any technical impacts since this is only use on the console. Some short lived user impacts of course when pressing `S` doesn't work as expected.

(2) Use the `S` mapping for the majority of (non S100) users, and for S100 user (2a) remap the Slice function to another key, or (2b) simply disable it.

Opinions

Option 2a - is the only option I dislike, it promotes both user and code confusion, where menu options can arbitrary mappings. Its a precedent I dont think should be set.

Option 1 - seems preferred to me, simple to implement and understand

wwarthen:
@kiwisincebirth, Let's go with option #1.  I will change the S100 Monitor invocation key.

Thanks, Wayne

kiwisincebirth:
> I haven't really reviewed invntslc.asm thoroughly, but wanted to get this merged. Any issues in invntslc.asm can be addressed later.

Noting The code was copied from slabel.asm so any issues will be shared.

> At some point, it looks like there should be some minor updates to the Boot Menu section of the Applications Guide.

I will look at this

kiwisincebirth:
[...]

PR #577, opened 2025-06-28
Added New Rom App - Slice Inventory - executed from RomLdr "S"

kiwisincebirth:
See Issue #575 

Dev Notes:

The changes involved primarily impacted RomLdr, primarily for usability. Rom Apps are launched with verbose "Loading..." messages, and typically shoutdown via a warm reboot.

For the "S" command wanted a streamlined UX, his necessitated 
* the RomApp itself perform a RET on completion
* A new `romcall` (romldr) function to load the app (based on `romload`) without the "Loading..." messages.
* And 2 core new functions shared by both `romcall` and `romload`
* new function `romcopy` which does the actual load of romapp without any messaging.
* new function `findcon` find app by console key A and return IX pointer to app

creating the above 2 function improved code reuse, since there are several mechanisms (DSKY,NVRAM,CONSOLE,FP-SWITCHES) where a rom app needs to be loaded, and each has a unique code path.


kiwisincebirth:
Note the windows scripts have not been tested.

PR #596, opened 2025-07-19
ch376-native: extended assign to support new device type usb

dinoboards:
Extended the `ASSIGN` app to support the new USB device type.

Did some basic testing - unassigned/re-assigning/listing etc of a USB device

As I am not familiar with the code -- I'm not sure if there maybe are other areas that may also need adjusting..

But at least nothing seemed to break with my testing.

Cheers
Dean


wwarthen:
Thanks for the quick turnaround on this @dinoboards.

Will review and merge shortly.

Thanks, Wayne

Issue #456, opened 2024-11-15
ch376 native usb support


dinoboards:
Hi Wayne,

This is not eZ80 related.  

Something else for your consideration.

I have been progressing over the last few weeks, the porting into RomWBW, the code I wrote for my MSX kits to drive the CH376 usb module.  This code enables full usb protocol access - enumerating hubs, and many devices.

My msx build has support built-in to drive USB hubs, floppy disks, mass storage devices, keyboards and printers - with plans for other usb device types in the future.

I had always wondered how i could port this code to RomWBW.  The first challenge is, I am using the `z88dk` c compiler to write the code.

I do have a working version on my local fork of RomWBW.  See image of the boot up messages:

![image](https://github.com/user-attachments/assets/5f996e94-6708-49c6-8b99-95109c0ed247)

> This is booting up using my ez80 -- but the code should work just fine on a standard Z80 CPU - as it does for my MSX system

To make this work, I had to create a 2 stage build process.  First, use z88dk c compiler to convert the c code to assembler.  Then load that assembly code into the HBIOS's standard build process.

I suspect you don't want to add z88dk to your toolchain requirements - and I would image it would be a challenge for window builds.

This is what i have done so far, for the build process.

* I have added the relevant c code to the directory `Source/HBIOS/ch376-native/source-doc`
* I have a makefile that when run, will compile/transpile the c code into assembly files in directories: 
  * `Source/HBIOS/ch376-native/base-drv` (common code to enumerate and communicate usb devices)
  * `Source/HBIOS/ch376-native/scsi-drv` (usb mass storage hbios driver)
  * `Source/HBIOS/ch376-native/ufi-drv` (usb floppy hbios driver)
  * `Source/HBIOS/ch376-native/keyboard` (usb keyboard hbios driver
* Added to hbios, new config switches to enable optional including of the drivers.   The HBIOS build process just see conventional assembly files to process. (so should work on windows etc)

At this stage, this means that if anyone pulled my fork, and does the normal HBIOS build process, they will be able to build in the usb drivers - no need to have z88dk installed etc.

But if someone wanted to 'patch' or change the code, they can of course change the 'generated' assembly.  But they really should change the C source files and regenerate the assembled code.

I am hesitating to create a PR - as I am not sure if this is something you would want - and understand you might want to review and think about this a bit.

[...]

PR #600, opened 2025-07-21
ch376-native: output the configured port numbers for the ch376 driver

dinoboards:
Include the configured IO ports in the boot log for the ch376 native usb driver.


wwarthen:
Thank you @dinoboards.  I am merging now, but I do have a question that I will post on the Issue #456.

PR #609, opened 2025-08-29
ch376native: fix issue with CH376S modules with VER 3 firmware.

dinoboards:
CH376 native/direct USB driver fixes:

1. Improved timing to detect module on cold boot (changed delay loops from 'medium' to 'short')
2. Fix issue with Version 3 of CH376 module firmware -- initial device enumeration request would return a USB STALL state. Added retry logic to clear the STALL and try again - only retries once.

C Code and new generated assembly code included in commit.


wwarthen:
Thank you @dinoboards!

-Wayne

Issue #625, opened 2025-10-19
Z80 assembler expression evaluation


b3rendsh:
After copying a large file (512KB) in CP/M 3 or ZPM3 the system will halt on MSX. 
--> the memory manager isn't using the correct memory segments in upper memory.

It wasn't easy to find the bug but it turns out the assembler isn't parsing expressions as expected.

There is following code:
	`LD	A,((ROMSIZE + RAMSIZE) / 16) - 2 + RAMBIAS * 2
`

With ROMSIZE=0, RAMSIZE=448 and RAMBIAS=2 I expected it would result in:
	`LD	A,30`

With the z80 assembler the order of operations however is from left to right:
((0+448)/16) - 2 + 2) * 2 = (28 - 2 + 2) * 2 = 56

Following code also won't be parsed as expected:
	`LD	A,((ROMSIZE + RAMSIZE) / 16) - 2 + (RAMBIAS * 2)`

It will result in:
	`LD	A,(30)`

A solution that works is to separate the expression from the instruction i.e.

```
RAMSEG  .EQU    ((ROMSIZE + RAMSIZE) / 16) - 2 + (RAMBIAS * 2)
        LD      A,RAMSEG
```

I'll fix the code in a new pull request, unless there's a better way to handle this.

wwarthen:
Yeah, I have been bitten by this multiple times.  Z80 assemblers have trouble with the LD A,(x) syntax because x mean load from an address or it can be an equation.

Your approach to this situation is good.  The other possible way to do it is to use something like LD A,0+(x).  This syntax will help the assembler understand that x is an equation.

Thanks, Wayne

wwarthen:
Definitely an issue with TASM.  It does have a warning somewhere in it's documentation about this, but something in the RomWBW docs would be nice.

Any thoughts on which of the RomWBW documents this warning should appear in?

Thanks, Wayne

b3rendsh:
In this document:  https://github.com/wwarthen/RomWBW/blob/master/CONTRIBUTING.md#coding-style

Issue #624, opened 2025-10-16
Year base and NVRAM usage in RP5C01 driver


b3rendsh:
See also discussion #613 

On MSX the 2-digit year is stored in the RTC relative to the base year 1980 e.g. 2025 is stored as 45. In RomWBW it is assumed to be relative to the base year 2000 e.g. 2025 is stored as 25.

In the RP5C01 driver this can be adapted for the MSX platform in the read/write time routines so it works the same.

Write routine:

```
	LD	B, REG_1YEAR
	LD	A, (RP5RTC_YR)
#IF (PLATFORM == PLT_MSX)
	; 2-DIGIT YEAR IS RELATIVE TO BASE YEAR 1980
	ADD	A,$20		; ADD OFFSET
	DAA				; BCD
#ENDIF
	CALL	RP5RTC_WRVL

```	
Read routine:

```
	LD	B, REG_1YEAR
	CALL	RP5RTC_RDVL
#IF (PLATFORM == PLT_MSX)
	; 2-DIGIT YEAR IS RELATIVE TO BASE YEAR 1980
	SUB	$20			; SUBSTRACT OFFSET
	DAA				; BCD
#ENDIF
	LD	(RP5RTC_YR), A

```

Maybe we can make this the standard behaviour for the RP5C01 driver if it's only used in MSX and the yellow MSX series,
or add a RomWBW configuration item  RTCYEAROFFSET (OFFSET IN BCD) that could also be used for other RTC drivers as well if needed.

Another thing is that the current NVRAM get/set routine will overwrite configuration data used by the MSX.
You can avoid this by using only bank 3 register (index) 1 to 12 and not use bank 2 where the MSX configuration is stored.
This will reduce the number of available NVRAM bytes from 12 to 6 (a RP5C01 register contains a nibble value).
Since some MSX games also use this free NVRAM storage area we can leave it as is for now and 
==> don't use the NVRAM bytes for RomWBW on MSX.

wwarthen:
Hi @b3rendsh,

I don't see any issue with implementing the way the year is stored.  It won't affect anyone using the HBIOS API.  @dinoboards, let us know if you have any concerns.

[...]

PR #626, opened 2025-10-20
MSX fix memory and base year

b3rendsh:
MSX fix upper memory segment assignment error and set RP5C01 base year to 1980.

See issue #625 and issue #624


wwarthen:
Thank you @b3rendsh

PR #640, opened 2025-12-14
N8PC default updates

danwerner21:
The 1.0 version of the N8PC changed the default Address of the PSG to better match the MSX and added Front Panel Switch support.  This updates the config to match.


wwarthen:
Looks good @danwerner21!

Thanks!

PR #646, opened 2026-01-08
HBIOS: added new config to enable support for Turbo Z80 CPU

dinoboards:
Hi @wwarthen 

It looks like you are in the middle of some refactoring/enhancements - so please ignore this PR as long as you want.  No rush - don't wont to add to your daily load.

This PR is a fix that has been an issue for my Turbo CPU module for a while now.  The issue is that SIO/2 blocks after one character has been sent to the CPU.

The cause is the `RETI` instruction is missed by the SIO/2 - and so it does not reset its interrupt flag.

Just a brief note, on what the Turbo CPU module is.

----

This is a Z80 RCBus/RC2014 module - much like any other - using a standard 40 PIN DIP Z80.  The Z80 runs 'normally' at the same clock as per other CPU modules (eg: ~7Mhz).  But unlike others, it will boost its clock to 20Mhz when its not doing any I/O operations.  This logic is driven by the IOREQ line of the Z80 - when that goes low, the CPU's clock is immediately slow to the bus clock for the next 31 clocks.  After that, it resumes at the full speed.

This means that many high cycle count instructions, such as the index instructions (IX, IY) are typically much improved.

----

So one compatibility issue that can occur, is when any peripheral device attempts to read the data bus, and see if the `RETI` instruction is executed - if the CPU is running at full speed - its likely not gonna see it.  So simple fix is to ensure CPU is running at non-boost speed when a `RETI` is executed.

And that is what this PR does - it adds a new config option `TURBOZ80` then when set to `TRUE` will issue a dummy i/o read operation just before the `RETI` instruction, forcing the CPU into its slow mode.  

Cheers
Dean


wwarthen:
Doesn't look like a problem to merge now.  The change is dormant unless specifically activated, so I'm not worried about it.

Thanks @dinoboards.

-Wayne

PR #647, opened 2026-01-08
Update loader for MSX

b3rendsh:
Minor update of the RomWBW loader for MSX: 

Added check if RomWBW is already loaded to prevent accidently running the program from RomWBW.
I also added a header text terminated by EOF (0x1A) so you can "TYPE MSX-LDR.COM" and get a small description of the program and the version, this used to be common practice for MSX / MS-DOS executables. 

See discussion #613


wwarthen:
Thanks @b3rendsh!

PR #656, opened 2026-02-02
tms: re-activated vram detection for V99x8 chips only

dinoboards:
Re-activate the vram detection for V99x8 only 

TMS probing faults as the memory is mirroed.  the code would need to be updated to determine if memory is unique or mirroed - but its probably not really needed

For TMS, we could perhaps assume 16K only.  I left this for a future change, if desired.

I dont have a F18A - so cant test that - so memory probe is bypassed for this chip.

Also, added constant literals for the VDP types to std


wwarthen:
> Minor note: the F18A is not MSX specific but a generic TMS chip replacement, works also in Coleco, TI99/4A, Nabu etc. Replace in std.asm `TMSVDP_F18A .EQU 4 ; MSX FPGA TMS9918 REPLACEMENT` with `TMSVDP_F18A .EQU 4 ; FPGA TMS9918 REPLACEMENT`

Thanks @b3rendsh.  I will update the comment when I do my reintegration.

PR #676, opened 2026-06-04
MM_Z180 support for HD64180

mmicko:
With Z180_RAMLOC we can now override RAM location, that is, by default, for Z180 at 0x80000, but on HD64180 address bus is 19 bit instead of 20, so RAM starts at 0x40000.
Added this value only in CFG_MASTER.cfg so it can be overwritten by other cfg only if needed.


wwarthen:
Thank you @mmicko.

-Wayne

PR #678, opened 2026-07-02
fix: zxcc 

wdl1908:
Hey, found a bug in Tools/unix/zxcc/zxcbdos.c - when running in a pipeline or docker non interactive without stdin it kept failing to run properly. This fixed the problem.


wwarthen:
Nice catch @wdl1908.  I will merge as soon as I can.

Thanks, Wayne

wwarthen:
Thanks @wdl1908.

PR #680, opened 2026-07-07
During experimentation with PCF and I2C communication I found some strange things.

wdl1908:
During experimentation with PCF and I2C communication I found some strange things. The PCF_WAIT_FOR_BB test was inverted and always hit the timeout. When changing that code to fix the bug, the DS7 driver started to act weird.  Investigating that revealed another bug. All fixes are in the commit, please check.
I also added PUSH BC and POP for PCF_WAIT_FOR_ACK for another project I'm working on code is almost ready but still needs some fine-tuning.


wwarthen:
Thank you @wdl1908.  I am currently out of town.  Will review and commit in a few days.

-Wayne

wwarthen:
I am not in a position to test these changes, but they look reasonable to me.  I am going to merge now.

Thank you @wdl1908.

-Wayne

PR #681, opened 2026-07-12
Fix building zxcc with modern GCC

grahamedgecombe:
GCC 15 defaults to C23. C23 assumes that a function declaration without an argument list has no arguments, rather than an unknown number of arguments, so it fails to build.

This PR adds argument lists to function declarations missing them.

No use of LLMs.


wwarthen:
Excellent!

Thank you @grahamedgecombe.

PR #683, opened 2026-07-21
Pcf quickfixes

wdl1908:
When doing some work on the pcf driver I noticed one glaring error the PRTSTR always return Z but that is in the ERROR path and should always return NZ
The second thing is the I2C FAILURE we discussed.


wwarthen:
Thank you @wdl1908.

-Wayne

PR #686, opened 2026-07-22
Pcf cleanup

wdl1908:
It took a while but now I'm confident in posting this.

These are 8 patches that should move ALL IN/OUT instructions out of the lcdi2c and ds7rtc and into pcf.asm in preparation for the bitbanged implementation. Every step was tested with lcd output confirmed and ds7 date testing done.

A>date
Wed 22/07/2026 21:34:41
A>date set

Enter today's date (DD/MM/YY): 27/07/26
Enter the time (HH:MM:SS):     21:34:50
Press any key to set time

A>date
Mon 27/07/2026 21:34:51

Which proves both directions work. 
Reading and writing to the ds1307 chip.

This was done after every patch and rebuild.
So I'm very confident that this is all working as expected.


wwarthen:
Thanks @wdl1908.

-Wayne

Issue #684, opened 2026-07-22
System (MSX) with NMOS CPU won't boot


b3rendsh:
One of the latest RomWBW changes causes MSX systems with a NMOS Z80 to crash in the boot process, near the beginning of the romldr module after HBIOS is initialized.

It looks like there's a logical error in the HB_PUSHINT / HB_POPINT combination:
HB_PUSHINT sets the carry flag if interrupts are disabled, so HB_POPINT should also skip the EI instruction if the carry flag is set. In the current situation if interrupts were disabled they are enabled and vice versa.

https://github.com/wwarthen/RomWBW/blob/512bd75868fa427e4fa248db781c50e77c07f8d8/Source/HBIOS/hbios.asm#L299
Changing the instruction `JR NC,$+3` to `JR C,$+3` will fix the error.

See also issue #620

wwarthen:
Hi @b3rendsh

> It looks like there's a logical error in the HB_PUSHINT / HB_POPINT combination: HB_PUSHINT sets the carry flag if interrupts are disabled, so HB_POPINT should also skip the EI instruction if the carry flag is set. In the current situation if interrupts were disabled they are enabled and vice versa.

Hmmm...  In the NMOS variant of HB_PUSHINT, the carry flag is supposed to be **set** if interrupts are **enabled**.  I think I see other problems though.  I think this:

`#DEFCONT \	JR	NZ,$+3`

should be

`#DEFCONT \	JR	NZ,$+4`

There is another problem.  The use of HB_DI in line 287 is problematic.  It may evaluate to `DI` or to nothing which will mess up the JR offset.  For now, I think `HB_DI` can be changed to just `DI`.

Let me know if this works.

Thanks, Wayne

wwarthen:
Sorry, my proposed solution does not make sense.  Now I don't know what the problem is.

wwarthen:
OK, sorry, I think you were right in the first place.  I will make this change unless you want to do a pull request.

Thanks, Wayne

b3rendsh:
No problem, I could have noticed the constant switching of interrupt status in the emulator debug screen. With the fix this has disappeared.
Created a PR for the change. 

/Henk

wwarthen:
Thanks @b3rendsh.  I will merge the PR now.

-Wayne

wwarthen:
I think this is resolved.  Closing the issue.

PR #687, opened 2026-07-23
fix error in nmos z80 pushint/popint macros

b3rendsh:
See issue #684


wwarthen:
Thanks @b3rendsh!

-Wayne

PR #688, opened 2026-07-23
fix: uz80as errors on out-of-range JR displacement instead of silently ignoring the error

wdl1908:
This one has bitten me 4 times in total, fixing it so other people don't fall in the same trap.


wwarthen:
Hi @wdl1908,

Nice catch.  I tend to do my work on Windows first, so I had not encountered the problem.  I checked the repository from which the RomWBW variant of uz80as was sourced, and the author also fixed this.  So, I am going to merge your update, but I am going to follow-up with the exact code that the author used so as to stay in-sync as much as possible.

Thanks! Wayne

wdl1908:
Ah I was under the impression the original project was death and did not check the upstream of that one. I was so fed up with encountering this thing multiple times that I had to fix it. Anyway thanks for the merge.

PR #693, opened 2026-08-01
fix: i2cscan PCF_WAIT_FOR_BB inverted busy/free polarity

wdl1908:
Small fix in an app programs to scan the i2c bus. should make it much faster to scan the bus.
Speed up from about 0.5 seconds for 16 scan addresses to about 50ms instead.
Effectively the condition was wrong it was nvested.


wwarthen:
Thank you @wdl1908!

-Wayne

PR #695, opened 2026-08-08
Fix I2CPCF clock-register verify to check the full written value

wdl1908:
Wayne,

Quick bugfix when changing pcf speeds.

Thanks Willy


wwarthen:
Thank you @wdl1908.

PR #696, opened 2026-08-09
Add dedicated DS1307 presence probe

wdl1908:
Wayne,

A big oversight in the code I switched pcf master controller board for a different board that had no ds1307 on board and found that the no presence detection did not work because it simple did not exist.

Here a patch to fix that.

Thanks Willy.


wwarthen:
Thank you @wdl1908.

Issue #694, opened 2026-08-02
New driver for I2c eeprom devices.


wdl1908:
Wayne,

Another saga in the I2C journey.
Most of the boards which have a DS1307 also have that little EEPROM of the 24Cxxx series or equivalent. It's a shame to not be able to use them from an app, and since we now have 2 master I2C drivers already that can be used interchangeably the idea was to add some kind of block device driver.

So I am working on one and I really wanted your feedback.
Looking through the DIO drivers in HBIOS (IDE, SCSI, USB, SD, etc.) they claim a unique numeric device-type ID from a shared table in hbios.inc.
I used $12 to identify the EEPROM as a block device. Looking through the docs there is an attribute field, but none of those flags apply to the EEPROM block device, we surely can't use them as driver letters in CP/M, so I set them effectively all to 0.

Two addressing modes, picked at compile-time:
1. Fixed I2C address for one EEPROM.
2. Scan the I2C bus from $50 to $57 for multiple EEPROMs (all the 24Cxxx share this address space)

Thanks. Willy

wwarthen:
Hi @wdl1908,

I had noticed the DS1307 modules tend to come with the 24C32 SRAM.  Certainly makes sense to have it accessible as a block device.

Yes, you would just assign the next available DIODEV device ID, which is in fact $12.  Just add an entry for it to the list of equates in hbios.inc.

Since the 24C32 is only 4K bytes of storage, I agree that it does not make sense to expose it as a drive letter in CP/M.  However, making it available as a raw block device under HBIOS is useful since this is non-volatile storage.  It could be very useful.

Unfortunately, the HBIOS DIO API does not have an ideal way to handle this type of "raw" block device.  At this point, CP/M is going to assign a drive letter to any DIO device that is "online".  Online means that the driver returns success (0) from the DIOMEDIA call.  I will probably need to make some minor adjustments to CBIOS, but we can deal with that later.

For now, I suggest the following:

- The media id byte returned by DIOMEDIA should return MID_NONE.
- The attribute byte returned by the DIODEVICE function should set bit 4 (LBA capable) and for bits 3-0 set to binary 0x5 (RAM).  The full byte value would be 0x15.

Scanning the I2C bus for multiple devices is fine, but will add complexity to the driver.

[...]

PR #698, opened 2026-08-14
I2ceeprom

wdl1908:
Wayne,

I've been working on this for a few weeks now It's related to the issue #694 we discussed.

Here are 3 patches 
First one to add the config to HBIOS.
Second patch is the real driver.
Third is the Makefile stuff to add Hitech-c in the build process and a test program to test the i2ceeprom driver.
If you do not like the makefiles stuff strip it out but this is what I like in my makefiles clear single line stuff for every action the makefile needs to take.
This has been tested against actual hardware with different rom sizes.

Anyway have fun with it.

Thanks Willy.


wwarthen:
Thank you @wdl1908.  Sounds good.  I am still really backlogged right now.  I will get to this as soon as I can.

-Wayne

wdl1908:
No worries Wayne. It's just me having fun with my childhood hobby again.

Thanks Willy.

wwarthen:
Thanks @wdl1908, this all looks good.  I think I will need to add the Windows part of the build fort this application, but that should not be hard.

Thanks, Wayne

PR #699, opened 2026-08-15
Various DMA fixes and improvements

grahamedgecombe:
This pull request contains a series of commits with some fixes and improvements for the DMA driver and monitor.

No use of AI.

I've tested this on my RC2014, both with DMAENABLE set and unset, to ensure I didn't introduce any regressions if the DMA is disabled.

I don't have any hardware for the ECB/MBC/DUO platforms to test against. However, the .com, .rom and .upd files for these platforms are bit for bit identical before and after my changes, so I'm confident I haven't broken the existing DMA support for these platforms.

Summary of the changes:

**Change default DMA port on RC2014/RCBus platforms**

On these platforms, change the default DMA port to 0x04 rather than 0xE0.

This is compatible with Fuzix, which uses port 0x04 for the DMA on RC2014 (https://codeberg.org/EtchedPixels/FUZIX/src/commit/90816b1cf3242da2b6cd9f65e7798871f0338aa3/Kernel/platform/platform-rc2014/kernel.def#L66).

I'm only aware of two DMA boards for RC2014 (https://groups.google.com/g/rc2014-z80/c/LYIp2YbLuiQ/m/MIab8ts3AgAJ and https://github.com/jonathanedgecombe/dma-module), and both of them allow the user to configure the I/O port with piano/slide switches, so at worst people using these boards in the wild will have to reconfigure the switches.

Furthermore, at the moment, the build breaks if you try to enable the DMA with the mode set to DMAMODE_RC, so it'd really surprise me if anybody was using RomWBW builds for RC2014 with DMA enabled anyway.

**Add DMAMODE_RC support to the DMA driver**

At the moment, if you try to build RomWBW with DMAMODE_RC, you get the following error:

```
dma.asm:59: undefined label (DMA_USEHALF)
```

This commit fixes that by actually adding support for DMAMODE_RC to the top of dma.asm.

The two RC2014 modules also do not have a separate hardware latch for controlling the /RDY input. This is fine, because RomWBW's DMA driver always uses the FORCE READY command, so the /RDY input is ignored anyway. This commit therefore also adds support for not defining DMA_CTL, and skipping writes to it if it isn't defined.

**Make DMA monitor platform configurations more consistent with HBIOS**

This adds DMAMODE_RC to the DMA monitor. It also restructures the #IFs so they mirror the ones in the dma.asm driver for consistency.

**Add support for platforms with no DMALATCH to the DMA monitor**

This is essentially the same as the DMA_CTL change in the DMA driver discussed above applied to the DMA monitor.

It is a larger change as the monitor uses this port in more places, so more of the code is compiled out on platforms without the configuration latch.

The monitor also always uses FORCE READY, so compiling this code out is not a problem for platforms without the configuration latch.

**Add DMAMODE_VDG support to the DMA driver and monitor**

I noticed that DMAMODE_VDG was defined but not actually used anywhere in the HBIOS.

I don't have a Velesoft Datagear to test, but I've added the DMAMODE_VDG configuration to the top of dma.asm based on the documentation/schematic (https://velesoft.speccy.cz/data-gear.htm).

It uses port 0x0B by default with a jumper allowing 0x6B to be set as an alternative, so I've also adjusted the default port used in the DMA monitor for it (previously it would've fallen back to 0xE0). Like the RC2014 modules, it doesn't have a configuration latch.

**Fix off-by-one error in memory to I/O port test**

The memory to I/O port copy test in the DMA monitor has an off-by-one error.

It's supposed to copy the string "0123456789abcdef", which is 16 bytes long, to the system's serial port. The DMA is programmed with a 16 byte length.

However, it actually ends up copying 17 bytes, because the Z80 DMA always copies one more byte than the programmed length.

This commit fixes that bug by programming a 15 byte length instead. I've verified the extra byte is no longer transferred with a logic analyzer.


wwarthen:
Hi Graham,

Thank you for submitting this PR.  They all sound like good changes.  I
wanted to apologize for my slow response; I have been sick for the last
couple days and I'm just now trying to start catching up.  Naturally,
GitHub is having system issues this morning, so I can't do any merges.  I
will take care of this as soon as possible.

Thanks, Waye

wwarthen:
Thank you @grahamedgecombe.  Everything looks good.

Truthfully, DMA does not get much use at this point, so I am not surprised there were issues.

Thanks, Wayne

PR #700, opened 2026-08-18
gitignore Hi-Tech C build artifacts (.obj, $$exec.$$$)

wdl1908:
Wayne more gitignores.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,11 +3,13 @@
 
 **/*.[Bb][Ii][Nn]
 **/*.[Cc][Oo][Mm]
+**/*.[Oo][Bb][Jj]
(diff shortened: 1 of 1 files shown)


wwarthen:
The global ignore of .obj is going to cause a problem with the .obj files in Source\Images\d_hitechc\u1

wdl1908:
Ah yes did not think about that one. I'll make a more specific one then.

wdl1908:
Wayne,

From what I can see the rule !Source/Images/** makes the images completely exempt from the specific *.obj rule.
So I think you can apply it like it is.

Regards, Willy.

wwarthen:
Right you are.  Merging now.

Thanks @wdl1908.

PR #701, opened 2026-08-22
Support multiple Hi-Tech C .com targets in one directory

wdl1908:
Support multiple Hi-Tech C .com targets in one directory
Fix failure detection with -s instead of -f


wwarthen:
Great, thank you @wdl1908!

-Wayne

PR #705, opened 2026-08-30
fix: skip DIO_ADDENT for a slot whose chip probe finds nothing

wdl1908:
fix: skip DIO_ADDENT for a slot whose chip probe finds nothing


wwarthen:
Looks good.  Thank you @wdl1908.

-Wayne

PR #707, opened 2026-09-04
Front Panel Adapter i2c002 and the software that goes with it.

wdl1908:
Wayne,

Another saga in the I2C journey. The i2clcd was only one part of the front panel changes; the other part is the LEDs and SWITCHES.

But for that I had to design some new hardware to adapt I2C to the Blue Box front panel switches made by Spencer Owen, the founder of RC2014.

Anyway, I added some documentation this time. If you don't like the Hardware addition I made, feel free to remove it.

The software is still the most important part to make the SWITCHES and LEDs work.

Thanks,
Willy


wwarthen:
Thanks @wdl1908.  This seems fine and I will merge.  I'm not in a position to test, so will trust that it works.

It does make me feel that I need to find a way to migrate the whole front panel stuff into some concept of driver files.  You have done what you could given the current HBIOS design, so this is fine for now.

Thanks, Wayne

wdl1908:
Yeah the front panel should really be it's own driver as this case exposes that there could be different drivers to handle it. but I did not want to go this far as it was only a few calls. But really that is your call to make.

PR #709, opened 2026-09-05
Add SAA1099 reset routine

flatduckrecords:
- add reset routine to the new SAA driver 
- change 'XOR' to 'XOR A' in sctim.asm which was preventing assembly with  uz80as


wwarthen:
Thank you @flatduckrecords.

-Wayne

PR #710, opened 2026-09-07
invntdev i2ceeprom label

wdl1908:
Wayne,

I hardly ever do cold boots, and when I do, I don't usually watch the boot messages. Two days ago I did, and noticed something strange:

```
Disk 7      Floppy Disk0:                                                                                                                                                                                                                                                               RAM Disk          256KB,LBA
Disk 8      Floppy Disk1:                                                                                                                                                                                                                                                               RAM Disk          256KB,LBA
Disk 9      Floppy Disk2:                                                                                                                                                                                                                                                               RAM Disk          128KB,LBA
```

After investigating, I added code to invntdev.asm to display these correctly, and set the attribute byte for the EEPROM device properly. Media type 13 was the cleanest unused value, so I picked that.

While diagnosing this, I also added a standalone CP/M build target for invntdev.asm (invntdev.com), so the drive listing can be checked without flashing a new ROM and cold-booting every time.

With this patch, the drive listing now shows correctly:

```
Disk 7      I2CEE0:     EEPROM            256KB,LBA
Disk 8      I2CEE1:     EEPROM            256KB,LBA
Disk 9      I2CEE2:     EEPROM            128KB,LBA
```

I also updated SystemGuide.md: added DIODEV_I2CEEPROM to the device type table, and documented media type codes 10-13 (SCSI, UFI, Other, EEPROM), which were already in use elsewhere but never listed. Only the EEPROM row line actually changed, but the whole device type table needed reformatting.

Willy.


wwarthen:
Very good @wdl1908.  Thank you.

-Wayne

PR #711, opened 2026-09-08
PCF8584 multi-master safety + dummy-read cleanup

wdl1908:
Wayne,

Another journey in the I2C Saga.

I was doing some work on the BF_I2C stuff that I promised to do and
found some problems. First I was really annoyed to have to switch
hardware between the T.Pycio board and the SC137, so I tried to put
them both on the back-plane and connect both to the I2C bus. Well,
that was an interesting experience, and it led me to these patches to
fix the multi-master problems.

Another problem I wanted to address in the same patch is the
dummy-read problem: the PCF8584 has a quirk where, once a read is
initiated, it needs to be read twice (don't ask me why "hardware"
does its own thing), which leaves it inconsistent with the i2cbit
driver, which doesn't have that problem. So in preparation for the
BF_I2C implementation, I fixed it so the i2cpcf driver does an
automatic double read of the byte, while i2cbit just does its own
thing and reads it normally. This brings both drivers to a point
where client software doesn't need to know which driver is
implemented on the backend.

While I was testing the multi-master stuff, I also rewrote some of the
scanner code to make it more portable and more correct in relation to
all the problems I found.

It occurred to me that people are sometimes lazy and don't want to
edit config or code to rebuild something, so I made sure they don't
really have to. All different versions are built now, no editing
needed unless it's real hardware addresses.

Also I included a simple utility to read the status byte of the
PCF8584, it was a very handy tool to diagnose problems.

During hardware testing, a number of other things surfaced and were
corrected. Tested all of this on the T.Pycio board with a genuine
second I2C master on the bus, the SC137.

Willy.


wwarthen:
Wow, quite a bit of stuff here @wdl1908.  It all sounds good.

- Certainly nice to have multi-master compatibility.
- Yes, the PCF8584 dummy read is messy.  Hiding this quirk in the driver is ideal.
- All of the other items also sound good.

A lot of files changed and I want to take the time to review them.  I will start on that now and merge when done.

Thanks! Wayne

wwarthen:
OK, looks good from a high level -- I am going to merge.  I think I need to add some stuff to the Windows build to sync up with the Linux build.  I will take care of that shortly.

Thanks, Wayne

PR #712, opened 2026-09-10
fix: bound I2C_GETBYTE_CYCLE wait, drain dummy byte before NACK

wdl1908:
Wayne,

I missed a crucial fix in the last merge request.

Willy.


wwarthen:
Thanks @wdl1908.

-Wayne

Issue #704, opened 2026-08-28
Possibility to make a real HAL for i2c called from CP/M


wdl1908:
Wayne,

Another saga in the I2C journey. This one is more involved.

I was trying to make some test programs for the LCD and other stuff I was working on, and decided to switch boards for the I2C master from the [Pycio board](https://github.com/tpycio/I2C-Module) to the [SC137](https://smallcomputercentral.com/rcbus/sc100-series/sc137-i2c-master-module-rc2014/). That resulted in me having to rewrite some of the base code to access the other board.

So I was thinking, would it be possible to have a real HAL in HBIOS for I2C, one that CP/M programs could use directly without needing a rewrite just because someone swaps in another board?

The underlying bus-master drivers in HBIOS should already be fine either way, I think that part is covered.

Where I get stuck is how something like this would actually be exposed. For i2ceeprom there's already a good solution via BF_DIO*, and that works great. But for general I2C I can't see how to start, since it's not a block device, so BF_DIO is out. It's sort of like a serial device, but not in the normal sense either.

Would it make sense to create a new call table, something like BF_I2C .EQU $60, and put all the I2C calls under that?

I'd like your advice on this, since you're the one who really knows HBIOS inside out, and you might be able to point me in the right direction.

Thanks,
Willy.

wwarthen:
Hi @wdl1908,

A very good question and one that I have mused on occasionally.

My original thinking was that I2C was simply an internal bus upon which other devices (disk, RTC, etc.) would interface.  As such, it would not really be exposed externally.  Analogous to SCSI bus devices.

With that said, I have seen that a few folks are interested in having access to I2C directly from userland.  I am sensitive to feature creep and bloat in RomWBW.  However, HBIOS is structured such that there is minimal overhead to adding an API set.  If there are no drivers, all that is added is a small dispatch routine.  In fact, even the dispatch routine could be optimized out in a build with no I2C devices.

So... I think I am OK with adding such an API set.  I don't personally have time to work on this, but you are probably quite capable of doing this.  I would have no problem with that if you are interested.

Just one warning if you choose to pursue this.  I am **much** more discerning about externally exposed interfaces than pure internal ones.  I would want to see a clear and complete definition of the API functions first.  I would probably have some opinions along the way.  😀

Let me know what you think.

Thanks, Wayne

wdl1908:
Wayne,

[...]

PR #714, opened 2026-09-14
Bf i2c hal

wdl1908:
Wayne,

Here it is the middle of the I2C saga.

Full HAL coverage and some test programs I wrote in C to really test if this was properly working.

I added some LIBR stuff in Makefile.inc to get around a limitation of zxcc command line length.

Be assured everything has been hardware tested with both I2C backends, I created special testing scripts to upload roms start them run bustest scripts and check the responses to make sure everything worked compared to a baseline.

I also included a build Variable to print the git commit id because I really lost track of what build worked where and which build was uploaded. If you don't like it just remove it from the build scripts and the sources but it helped me tremendously in debugging sessions.

Any remark, correction, addition let me know.

By the way this is the phase 0.5 of my plan phase 0 was everything that came before it just to get here.

I also included a simple i2cscan program written in C to show off the universal nature of the driver.

Well on to phase 1 to rewrite all the internal i2c drivers to use BF_I2C.

Have fun with it. I'm still having fun with it. There will be more I2c Sagas coming.

Willy.


wwarthen:
Hi @wdl1908,

Impressive progress.

The only thing I am not fond of is that the I2C functions are in HBIOS.  I see that you have the code protected by a conditional so it won't take space unless I2C is active, so that is good.  These days I am trying to move source code out of hbios.asm as much as possible because it has gotten far too big.

What if you moved the I2C functions and function table into an include file that is included by both drivers?  That way it would automatically be included if either I2C driver is enabled and the source code would not live in hbios.asm.

I should be clear that I am not talking about I2C_DISPATCH, just the function table and the functions.  I2C_DISPATCH clearly belongs in hbios.asm.

Thanks, Wayne

wdl1908:
Wayne,

Thank you for the praise, I've been working hard on this one.

Should I cancel this Merge Request and redo it, or when this is merged, redo the split in a separate file?

I would rather not include that in the I2C drivers as I have other plans for them.

An include in hbios should be doable.

Forgot to link it #704 

Willy

wwarthen:
Hi @wdl1908,

I will go ahead and merge the current PR since you have agreed to move the subject code to an include file.  I will assume that will be  part of a future PR.

I am not sure why the include file cannot be part of the drivers, but I am happy to wait and see how things shape up.

Thanks for all the excellent work!

-Wayne

wwarthen:
Hi @wdl1908,

So, a small follow-up on the git commands in the HiTechI2C Build and Makefile.  I like the functionality, but it is problematic for anyone building outside of a full git context.  For example, if downloading the source, the build will fail as a dirty tree.  It also requires that git be installed.

I know that as Windows/Linux developers this seems trivial.  But I can tell you emphatically that most RomWBW users will give up on this requirement.  They are not Windows/Linux developers.  I worked very hard to ensure that RomWBW would fully build without downloading any tools or setup under Windows.  Under Linux, there are some requirements, but minimal and installing packages under Linux is basically something all users need to be able to do.

At present, things are OK because HiTechI2C is not included in the build (it requires a standalone build invocation).  So, the build will not fail, but of course HiTechI2C will not be built.

If HiTechI2C is ultimately added to the default project build, it will be necessary to remove the git dependency.  I'm sorry about this, but it is a fundamental philosophy of the project.

Thanks, Wayne

wdl1908:
Wayne,

I will see if I can try to fix all your objections.
1. git not installed no problem sensible default.
2. not a git repo again sensible default.
same for build.cmd

Let me see If I can fix this for you.

Willy.

PR #717, opened 2026-09-20
fix: skip I2CEEPROM probe when the I2C bus master is absent

wdl1908:
During some testing I noticed that the bus test was missing.


wwarthen:
Thank you @wdl1908.  I'm out of town, so I may not merge till Thursday.

Thanks, Wayne

wwarthen:
Thanks @wdl1908.  Merging now.

-Wayne

